Issue #2, opened 2017-01-11
lintian issues with mars-sim-3.1.0-preview-5


shirishag75:
Lintian is a debian package-checker and does small checks to ensure that basic things are right. 

The following things were found to be an issue by lintian -

```
> lintian mars-sim-main-3.1.0-preview5.deb                                                                                   [95%]
W: mars-sim-main: changelog-not-compressed-with-max-compression changelog.gz
W: mars-sim-main: syntax-error-in-debian-changelog line 6 "found change data where expected next heading or eof"
W: mars-sim-main: latest-debian-changelog-entry-without-new-version
W: mars-sim-main: package-contains-timestamped-gzip usr/share/doc/mars-sim-main/changelog.gz
W: mars-sim-main: package-contains-timestamped-gzip usr/share/man/man1/mars-sim.1.gz
E: mars-sim-main: manpage-not-compressed-with-max-compression usr/share/man/man1/mars-sim.1.gz
```
 max-compression could be done by using gzip -9 

As far as time-stamping is concerned, that could be avoided by doing gzip -n when generating gzip. 

How time-stamps are bad can be known more at https://wiki.debian.org/ReproducibleBuilds 

changelog.gz itself has only old version data in it -

```
/usr/share/doc/mars-sim-main> zless changelog.gz   

mars-sim-main (3.1.0) unstable; urgency=low

  * Version 3.1.0 Debian package release.
 -- Manny Kung <mkung@users.sourceforge.net>  Wed, 21 Sep 2016 09:50:40 -0700

  * Initial debian package release.
 -- Scott Davis <scud1@users.sourceforge.net>  Wed, 31 Jul 2013 14:04:40 -0700

changelog.gz (END)
```
So at the very least, these need to be fixed.

mokun:
Sure. I'll do a max-compression and update the timestamp before the next preview version. Thanks for pointing that out.

mokun:
Question: is the time stamp supposed to be manually or automatically inserted ? 

I'm following the standard in [https://www.debian.org/doc/debian-policy/ch-source.html#s-dpkgchangelog] (https://www.debian.org/doc/debian-policy/ch-source.html#s-dpkgchangelog) when it comes to defining the time stamp.

For now, I'm rewriting the changelog as follows: 

> mars-sim-main (3.1.0-p6) unstable; Urgency : low
> 
>   * Version 3.07 debian package release.
[...]

PR #24, opened 2017-02-22
Remove embedded timestamps from changelog.gz and mars-sim.1.gz.

lamby:
gzip files can contain an embedded timestamp which can be omitted by
passing "-n". This prevents spurious Debian lintian(1) warnings which are
intended to aid development of reproducible packages.

(These are actually false positives for mars-sim as these files are
hardcoded and thus cannot vary between builds.)

Changes made via:

  $ gzip -9n < ./mars-sim-main/src/deb/changelog > ./mars-sim-main/src/deb/changelog.gz
  $ gzip -9n < ./mars-sim-main/src/deb/man/mars-sim.1 > ./mars-sim-main/src/deb/man/mars-sim.1.gz

Signed-off-by: Chris Lamb <chris@chris-lamb.co.uk>


lamby:
Related to #2

mokun:
sure! I'll use my fedora to compress both files with the -9n switch and upload them to svn.

larsnaesbye:
We'd need to merge it into the SVN repo I guess. That's the pain of not having the newest code here.

lamby:
Thanks :)

PR #379, opened 2021-07-20
Update JFreeChart to 1.5.3

larsnaesbye:
1.5.2 fixes bugs
1.5.3 add a plot type for Sankey diagrams

### DIFF
--- a/mars-sim-ui/pom.xml
+++ b/mars-sim-ui/pom.xml
@@ -47,7 +47,7 @@ http://maven.apache.org/xsd/maven-4.0.0.xsd">
 		<dependency>
 			<groupId>org.jfree</groupId>
 			<artifactId>jfreechart</artifactId>
-			<version>1.5.1</version>
+			<version>1.5.3</version>
 		</dependency>
 
 		<!-- trident is required by steelseries.extras in mars-sim-ui		


mokun:
> Sankey diagrams

Never heard of the term "Sankey diagram" until you mentioned it. I just looked at https://en.wikipedia.org/wiki/Sankey_diagram and it certainly looks cool for illustration.

Where do you see in mars-sim we can give it a shot ?

larsnaesbye:
It's apparently often used for energy flows, and maybe there is no obvious place where we could use one right now. I just thought it was an interesting add.

larsnaesbye:
Also of interest is that they've now deprecated the PieChart3D because they think we should their other chart library https://github.com/jfree/orson-charts instead

mokun:
> Also of interest is that they've now deprecated the PieChart3D because they think we should their other chart library https://github.com/jfree/orson-charts instead

Okay. 

Guess they just made orson-charts compatible with Swing.

We can try converting our codes to using orson charts.

mokun:
I just checked out this [page](https://github.com/jfree/jfreesvg).

This [Sankey diagram](https://raw.githubusercontent.com/jfree/jfreesvg/master/src/main/javadoc/doc-files/FlowPlotDemo2.svg) looks really nice.

larsnaesbye:
> Guess they just made orson-charts compatible with Swing.
> 
> We can try converting our codes to using orson charts.

We could indeed. But maybe it's not high priority....

PR #402, opened 2021-07-28
Codesmells part2

bevans2000:
Removed some more serious code problems plus minor errors in the package.html


mokun (approved):
those html pages are auto-generated whenever we run the HelpGenerator to create a set of html data files for the GuideWindow.

So we actually need to make sure we update HelpGenerator to create the html compliant header.

bevans2000:
> those html pages are auto-generated whenever we run the HelpGenerator to create a set of html data files for the GuideWindow.
> 
> So we actually need to make sure we update HelpGenerator to create the html compliant header.
No these are the JavaDoc package.html files that have been manually created. But I will update the HelpGenerator

Issue #409, opened 2021-08-04
Remove the hardcode Reporting Authorities


bevans2000:
**Describe the improvement**
The Reporting Authorities are hard coded within the `ReportingAuthorityType` enum. With the development of the 'governance.xml'  (#363) it is now possible to remove the use of the enum altogether and use the externally defined Reporting Authorities.

PR #413, opened 2021-08-04
Reporting Authority range is defined totaly by the governance XML

bevans2000:
I tis now possible to add a new ReportingAuthority & Country to the governance XML file and it will be used by the Simulation. No code changes are required.
If there are no Vehicles names defined then a default <Vehicle type># convention will be used.
If a Country is referenced where there are no names defined then it defaults to USA as that has a large name range. This could be changed to random in the future maybe.

#409


bevans2000:
So I have cleaned all my branches and retested. The GUI comes up in all cases as below:
- Feature branch in my fork
- master branch in my fork
- master branch in mars-sim

Not sure what it could be. I have previously seen the early pulses go into an infinite loop. These are driven by the random values so do not always occur. For example the previous bug with the Crop class on large Greenhouses.
 
Also the UI starts in a very odd way because it is built inside the AWTEvent thread via the Swing.invokeLater; this has a chance to create a deadlock in the JVM monitors if it clashes with the Console. I'm sure it is not this problem but does make it difficult to debug

mokun:
Hmmm. I don't know what happened. 

In eclipse, I'm not able to load up main window.

image

The main window is white blank in the task bar on windows 10 and cannot even be focused on and brought up.

I've cleaned and rebuilt twice in Eclipse the result is the same

mokun:
The same issue in IntelliJ.

image

mokun:
are you using linux ?

bevans2000:
One thought. Have you tried removing your ui_defaults before starting?

bevans2000:
Things to try:
- Does the site editor run and will the ui start via that route? This will have a difference initialisation.
- Remove the ui defaults file
- Run it in the debugger to see if any exceptions are being thrown and hidden by the ui code
I'm busy with work the rest of this week but I will try and create a single threaded version of the ui startup on a branch to see if that fixes it.  Also how many cores do you have? I can spin up a Windows machine in AWS close to yours and see if that shows anything different to mine. I'll also try my Linux laptop

mokun:
> * Remove the ui defaults file

oic. After I remove it, it works now. 

Guess I plug and unplug the second monitor earlier and it affects where the main window is saved at.

bevans2000:
> > * Remove the ui defaults file
> 
> oic. After I remove it, it works now.
> 
> Guess I plug and unplug the second monitor earlier and it affects where the main window is saved at.

Good news. We need to make that logic a bit more defensive.

mokun:
> make that logic a bit more defensive.

Yea in JavaFX, it's easier to check for the presence of second screen. But swing should have something to allow for that as well.

PR #421, opened 2021-08-20
Support for multiple Scenarios

### LINKED ISSUE #411: Create a Scenario concept by isolating the initial Settlements
Separate the initial settlement details from the settlments.xml file into a scenario.xml. "Fixed" aspects such as templates will stay in the settlments.xml.
Scenario will be:
- Have a name & description
- Defines a set of Initial Settlements & Arriving Settlements
- Have a `Scenario` class within the code base to encompass the initial set-up
- Site Editor modified to the allow saving of Scenarios
- Support command line & Console options to select a saved Scenario
-  Have an export feature to allow Scenarios to be shared

bevans2000:
The Simulation Config Editor can now saved and reload different Scenarios. The controls are the same as the CrewEditor as they have been consolidated around the concept of a `UserConfigurable` item. Eventually Reporting Authority shoudl also use this interface with an new associated editor.

Scenarios can be create or copy from the default one and saved. These are held as XML files on disk.

There is also a new Founders crew pre-defined that contains contributors to the GitHub project.
close #411

### DIFF
--- a/mars-sim-core/src/main/java/org/mars_sim/msp/core/SimulationBuilder.java
+++ b/mars-sim-core/src/main/java/org/mars_sim/msp/core/SimulationBuilder.java
@@ -16,6 +16,8 @@ import java.util.logging.Logger;
 
 import org.apache.commons.cli.CommandLine;
 import org.apache.commons.cli.Option;
+import org.mars_sim.msp.core.configuration.Scenario;
+import org.mars_sim.msp.core.configuration.ScenarioConfig;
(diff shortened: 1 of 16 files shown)


mokun:
> There is also a new Founders crew pre-defined that contains contributors to the GitHub project.

Oh do you mean that list of real people names that are associated with the Mars Society in people.xml ? 

yea guess we may as well call all of them founders. 

May be we should give them some special skills or bonus abilities so that they are welcome and invited to other settlements to kick start special projects ?!

PR #445, opened 2021-09-25
Redesign Vehicle Loading

bevans2000:
The main change of this Pull Request is a change to how Vehicles are loaded. Now there is a `LoadingController` class that is owned by a `VehicleMission` and shared by all loading Tasks for that Mission. The controllers represents a loading plan where it holds a manifest of items needed which is taken at creation time from the VehicleMission. As items are loaded the manifest is reduced; hence once the manifest is zero then the vehicle is loaded.
This greatly improves the simplicity as all logic is in one place since the load garage & load EVS task now delegate most logic to the controller. Also there is now no need to continuous check the inventory in the Vehicle.
A secondary change as result of debugging the change are some Console changes:
- The Mission Summary command shows the associated Vehicle
- The mission command now shows the remaining waypoints for a Mission as well as the current manifest of the loading plan
- The vehicle crew command only show the occupants when in a Settlement; previously it showed the Mission Member. When outside a Settlement the crew still shows the mission members. This ensure the crew member doing exploration at a site are still show as Crew of the Vehicle.

There is a small further outstanding change relating to the VehicleMission.isVehicleLoadable method which is still called too often during the loading. This needs a change to the mission logic since this predicate shoudl eb checked in the mission constructor and then trusted for the course of the mission.

Also I suspect there is something not correct with the probably score of `Delivery` missions as they are nearly always zero.


mokun:
How do I clear this conflict ?

```
		try {
			List<Future<String>> results = executor.invokeAll(settlementTaskList);
			for (Future<String> future : results) {
				future.get();
			};
		} 
		catch (ExecutionException ee) {
			// Problem running the pulse
            logger.log(Level.SEVERE, "Problem running the pulse : " + ee.getMessage(), ee);
<<<<<<< loading-issue439
=======
            // Note: Use printStackTrace for debugging only. Must comment out printStackTrace 
            // or else sonarcloud gives an "E" grade.
            //ee.printStackTrace();
>>>>>>> master
		}
		catch (InterruptedException ie) {
```

In my last commit, I added a comment statement on UnitManager saying this

> // Note: Use printStackTrace for debugging only. Must comment out printStackTrace 
> // or else sonarcloud gives an "E" grade.

It won't allow me to merge the pull.

When I click on `Resolve conflicts` button, there is no options available for me to choose.

mokun:
> When I click on `Resolve conflicts` button, there is no options available for me to choose.

oh guess it's just waiting for me to manually EDIT the UnitManager and remove the highlighted conflicts.

it's done!

bevans2000:
There was no conflict when k created it. I guess you did a commit in the meantime

bevans2000:
Sorry about that @mokun . You should have pushed the request back to me for me to merge it. That is normal practice because it's my problem there was a conflict.

bevans2000:
> How do I clear this conflict ?
> 
> ```
> 		try {
> 			List<Future<String>> results = executor.invokeAll(settlementTaskList);
> 			for (Future<String> future : results) {
> 				future.get();
> 			};
> 		} 
> 		catch (ExecutionException ee) {
> 			// Problem running the pulse
>             logger.log(Level.SEVERE, "Problem running the pulse : " + ee.getMessage(), ee);
> <<<<<<< loading-issue439
> =======
>             // Note: Use printStackTrace for debugging only. Must comment out printStackTrace 
>             // or else sonarcloud gives an "E" grade.
>             //ee.printStackTrace();
> >>>>>>> master
> 		}
> 		catch (InterruptedException ie) {
> ```
> 
> In my last commit, I added a comment statement on UnitManager saying this
> 
> > // Note: Use printStackTrace for debugging only. Must comment out printStackTrace
> > // or else sonarcloud gives an "E" grade.
> 
> It won't allow me to merge the pull.
> 
> When I click on `Resolve conflicts` button, there is no options available for me to choose.

Yes Sonar Qube us correct. So I've changed it to pass the Throwable to the logger. The logger will do the stack trace output but not to standard out so that avoids the flag.

mokun:
> That is normal practice because it's my problem there was a conflict.

oic

mokun:
> So I've changed it to pass the Throwable to the logger. The logger will do the stack trace output but not to standard out so that avoids the flag.

So is what we have currently working well to both show stack trace and avoid sonarcube flag ?

bevans2000:
> > So I've changed it to pass the Throwable to the logger. The logger will do the stack trace output but not to standard out so that avoids the flag.
> 
> So is what we have currently working well to both show stack trace and avoid sonarcube flag ?

Yes should  be. I changed it help with the redesign. Also I changed how the time pulse is applied to children Units in Settlement. This will minimise the impact of affecting other Units.

bevans2000:
There are a 7 CodeSmells with the Pull Request but 4 were there already. I'll sort them out on the next commit I do.

PR #489, opened 2021-10-31
Optimise Crop calculation

### LINKED ISSUE #467: Tending Crop time
**Describe the bug/issue**
- Tending crop time is not valid at the end of the harvest

**Actual/Observed Behaviors**
```
00-Adir-03:059.014 (Info) Fishery : [Port Braun] Large Greenhouse 1 - Weeds fully tended for 54.0 millisols.
00-Adir-03:069.651 (Info) PlanMission : [Port Braun] Ashley Harris - Looking into the settlement's mission needs.
00-Adir-03:069.651 (Info) PlanMission : [Port Braun] Ashley Harris - Determined that the settlement doesn't need a new mission.
00-Adir-03:072.676 (Info) Crop : [Port Braun - Large Greenhouse 1] Daniel King - Harvested a total of 102.99 kg Strawberry.
00-Adir-03:072.676 (Info) Crop : [Port Braun - Large Greenhouse 1] Daniel King - Closed out the final harvest of Strawberry.
00-Adir-03:072.676 (Warning) Mind : [Port Braun] Daniel King - 'Tending Crops' returned a remaining time 17848.68359017087 larger than the original 0.04037161531203451
00-Adir-03:073.359 (Info) DigLocalRegolith : [Port Braun] Terry Allen - Checking in 43.4 kg regolith.
00-Adir-03:073.835 (Info) Farming : [Port Braun] Large Greenhouse 1 - 0.35 kg potato tissue-culture was fully available.
00-Adir-03:073.835 (Info) Crop : [Port Braun] Large Greenhouse 1 - Done growingPotato's tissue-culture. Transferring its plantflets to the field.
00-Adir-03:073.835 (Info) Farming : [Port Braun - Large Greenhouse 1] Ashley Harris - Planted a new crop of potato.
```

**Specifications**
- r6275

### LINKED ISSUE #485: Optimise Crop calculations
**Describe the improvement**
The ```Crop``` class duplicates a lot of the ```CropType``` properties which due to the large number of Crop instance will consume excessive memory. Secondly there are numerous lookups against the ```CropConfig``` to retrieve details.

There can be optimisations to improve the logic. Due to the large number of Crop instances any improvement could have a large improvement on CPU cycles.

bevans2000:
Some of the shared calculations have been moved to the ```Farming``` class. 
There is a reduction in memory consumption as the ```Crop``` no longer duplicates properties of the associated ```CropType```. The Crop logic has been changed to be based on ```PhaseType``` instead of the index in the phase list. This make the code more understandable and better supports different crop phase cycles, e.g. Copt phases no longer have to have a Maturation phase at position end-2.

close #485 
close #467


mokun:
> e.g. Copt phases no longer have to have a Maturation phase at position end-2.

So which `CropCategoryType` or where did you say you eliminate *Maturation* `PhaseType` ?

bevans2000:
So I've changed nothing. What I have done removes the need for the Maturation phase to always be at position end - 2. As the logic is PhaseType based we can  now have different growing patterns for the Crop Categories. Also we don't need to always have a Planting & Incubation phase.

mokun:
Sure. I'm thinking of making the time for harvesting crop less rigid because, in reality, there's no such thing as exactly 100% harvest ready. 

If the population is hungry, they can harvest a crop during the entire last phase, say starting 80% when one can start harvesting.

mokun:
To make a food crop ready for harvest, we can create a standard harvest rate for that crop category.

Then we can use a check box and a number spinner box to let player to decide if they want to use the default harvest rate or roll in their own rate.

Say, if  carrot has a default harvest rate of 1. Player wants his own rate and set it to 1.5. Then it will be faster to harvest the carrot. But the harvest amount would be shrunk from, say, 40 kg to 27 kg.

mokun:
Btw, we can rename the class `CropType` to `CropSpec`, as you have suggested, since by mars-sim coding convention, `___Type` is more like a enum class.

mokun:
Also, in terms of harvest, if the crop is reasonably healthy, settlers don't have to uproot everything after the harvest is over, especially for non-vegetable types.

We could allow some part of the plant to exist. Thus it doesn't need to start from zero percent. Rather, it can start from, say, 10%, depending on when the 2nd or 3rd `PhaseType` begins. 

So, in future, if we add Tree types, such as apple, orange, lemon, etc., it makes more sense to simulate the harvest of fruits. The bulk of the tree with its root system would remain intact after the harvest. This would add meaning to the percentage bar, making it to mean more like the harvest index readiness.

bevans2000:
> Btw, we can rename the class `CropType` to `CropSpec`, as you have suggested, since by mars-sim coding convention, `___Type` is more like a enum class.

Sure, good point.

bevans2000:
> Also, in terms of harvest, if the crop is reasonably healthy, settlers don't have to uproot everything after the harvest is over, especially for non-vegetable types.
> 
> We could allow some part of the plant to exist. Thus it doesn't need to start from zero percent. Rather, it can start from, say, 10%, depending on when the 2nd or 3rd `PhaseType` begins.
> 
> So, in future, if we add Tree types, such as apple, orange, lemon, etc., we can continue to use Farming to simulate the harvest of fruits. The bulk of the tree with its root system would remain intact after the harvest. This would add meaning to the percentage bar, making it to mean more like the harvest index readiness.

We could maybe do this by not having a FINISHED phase so it last forever, bad health being the only thing that stops it. I'll have a think.

Issue #470, opened 2021-10-16
Create LocalPosition class for Units


bevans2000:
**Describe the bug**
Many Units and classes use the X, Y, Z as separate properties when calculating Walking paths and current position in Buildings.
This is inefficient and means 3 parameters or 3 set of getter/Setters when a single one would be better.
This also clashes with the Buildings/Activity sports which are defined in terms of Point2D. This is latter class is actually an AWT class and is quite heavyweight compared to a simpler version of our own. Plus the Point2D is not serializable itself but only it's private subclasses.
We should replace both areas with our own class that is integer based to make the arithmetic quicker. The position units should therefore become centimetres instead of meters.

mokun:
Sure, we can create Location class with x, y, z when a person/robot is inside a vehicle or a settlement.

Once a person gets outside the settlement, it will switch to Coordinates class.

mokun:
If we're switching over to javafx for UI, we may as well use `javafx.geometry.Point2D` class and roll in our own Line2D and Area for walking and collision. See https://stackoverflow.com/questions/32475229/which-point2d-should-i-use.

bevans2000:
Our use of a 2d point in not about the UI so we should use our representation.

bevans2000:
Also if the external class is extended to do other things we may be storing objects that are bugger than we need. I would like to

1. Replace the use of Point2D with our own simple POJO in the Activity spots. Plus make the lists immutable once constructed so they can be safely shared and remove any local copying
2. Combine all uses of x & y arguments to a single POJO; plus removing the X & Y from Person & Robot
3. Convert our own position class to be integer based instead of double. This will speed up any arithmetic and reduce storage, We would use centimeters as the unit instead of the current meter. A singe centimetre is accurate enough for our purposes.

bevans2000:
Closed by accident in the pull request.
Next is activity spots followed by Walking Tasks but with some optimisation

bevans2000:
Found a great optimisation during this reworking concerning ```WalkingSteps```. Currently the pattern we have is

1. Call ```Walk.canWalkSteps()```
2. Which in turn creates a ```WalkingSteps``` instance of the route.
3. If it return true then we create a ```Walk``` Task
4. The Constructor then creates another ```WalkingSteps``` instance with exactly the same parameters.
5. The constructor validates the 2nd Walking steps instance.

[...]

PR #523, opened 2021-12-06
Apply LocalPosition to BuiildingAirlock & WalkingSteps

bevans2000:
This is a change to replace Poitn2D with LocalPosition in the Walk logic. This has resulted in:

- BuildingAirlock logic to calculate the various activity positions has been simplified.
- No need to main a list of activity positions and map as the Map can cover both purposes
- Optimised the Walk logic. Previously 2 WalkingSteps instances were created and evaluated for each Walk task. Now only a single one is created.
- Walk class has factory methods to encapsulate the WalkingStep class away from the caller
- Remove the need for Robots to have a separate WalkStep class.
- Reactivated all the tests of the WalkingStep Unittest. Created some factory methods to simplify the tests and allow easier creation of new tests.

This Pull Request will allow the next one to replace the Point2D in the ActivitySpots which is a major step forward. And also remove the deprecated method calls that have been deliberately created as part of #470.


mokun (approved):
Good changes !

bevans2000:
> Good changes !

Let's hope so....  Ran lots of tests..

Issue #524, opened 2021-12-07
Add a Settlement Combo Box, Building Tab in Monitor Tool


mokun:
**Is your feature request related to a problem? Please describe.**
- Currently, all robots, crops, mission, goods, etc. have their own respective tabs in Mission Tool. 
- However, each tab show the units across all settlements. 
- Even though a player can sort the settlement column so as to highlight what units are in each settlement, this is considered INFORMATION OVERLOADING.

![image](https://user-images.githubusercontent.com/1168584/145088990-eefdc41f-7423-4c34-8372-daad9717449f.png)

**Describe the solution you'd like**
- Have a settlement combo box to choose which settlement's unit to view in Monitor Tool.
- Should be able to choose one settlement at a time but may also choose all settlements under one's jurisdiction/sponsorship.

mokun:
After the last two commits, it looks like this now:

![image](https://user-images.githubusercontent.com/1168584/145144836-3c58dde6-bf38-4044-9059-925556c9d627.png)

bevans2000:
@mokun I like this feature. Should the Settlement tab still should ALL settlements?
Also I think we should add a Building tab and remove the building tab from the Settlement Unit window.. Then selecting a row in the Building tab would display a new Building Unit Window.

mokun:
> Should the Settlement tab still should ALL settlements? 

hmm you point out what the settlement tab should have.

Now, I have in something else in mind.

What kind of visibility should sponsor A have over settlements that are not his own ?

So, in following this train of thought, we need a *sponsor* mode, not a *commander* mode.

A *sponsor* mode is for a player to play the role of a sponsor to manage a list of his settlements.

mokun:
> The Crops tab does not react to the Settlement combo.

Right. We need figure out what useful info we should place in the crop tab.

For now, it's just not important enough to show what category of food crop are being planted....

There are many concerns regarding how well a greenhouse should operate to maximize the yield and minimize the inefficiencies.

None of these are captured in the crop tab so far.

What do you have in mind ?

mokun:
>  should add a Building tab and remove the building tab from the Settlement Unit window.. 

> that would be interesting to show what's going on in a building. The atmosphere, power, the thermal status, etc.

> Then selecting a row in the Building tab would display a new Building Unit Window.

Do you have a visual example ?

bevans2000:
> A _sponsor_ mode is for a player to play the role of a sponsor to manage a list of his settlements.

[...]

PR #529, opened 2021-12-15
Created a Building tab to the MonitorWindow

bevans2000:
This requests adds a new tab holding Buildings to the MonitorWindow. Th new tab reacts to the Settlement filter.

Selected rows will be displayed in the rebuilt BuildingWindow. This window has a tab per BuildingFunction. These tabs are seeded from the panels in the Settlement Building tab. These panels have been reworked to provide a more consistent approach.
The Building tab has been removed from the Settlement unit window.

#524


mokun:
@bevans2000 

Can you resolve that ? 

I happened to add the ability in PopUpUnitMenu to view construction sites.

It seems like just the conflict with import statement

mokun:
not just import statements.

A few other places regarding the recognition of construction sites

mokun:
still having conflicts...

bevans2000:
Sure. There were no conflicts when I created it so it must  the the recent commits. I will merge the master and do it that way.

bevans2000:
I've merged it but the "quickStart" option is generating NPE and no UI starts. I will have a look but it seems the UI is constructed out of sequence now. I'll fix it.

PR #530, opened 2021-12-16
Fixed problem with NPE in quickstart mode

bevans2000:
Root cause is UI classes are hold invalid static references to
Simulation. Generally status references should not be initialised as
part of class loading unless the references instances themselves are
static and unchanged.

Changes:
- MainWindow & MainWindowPane hold instance reference to Simulation not
static. Simulation instance passed in constructor
- SimulationBuilder start method returns the created Simulation
- MarsProject passes the newly created Simulation to the UI as part of
the start-up
- SimulationConfig is being double loaded again; added the loaded flag
back to Config class
- Added '-cleanui' method that will force the UI to be clean and ignore
the saved UI configuration
- Having the '-new' will now bypass the user questions in the console
- AudioPlayer gets the Simulation from the MainDesktopPane (parent)
instead of holding a static reference
- Added a new Eclipse Configuration that is a vanilla start taking no
arguments

### DIFF
--- a/mars-sim-core/src/main/java/org/mars_sim/msp/core/SimulationBuilder.java
+++ b/mars-sim-core/src/main/java/org/mars_sim/msp/core/SimulationBuilder.java
@@ -242,8 +242,9 @@ public class SimulationBuilder {
 
 	/**
 	 * Uses the previously defines options and start the required Simulation
+	 * @return The new simulation started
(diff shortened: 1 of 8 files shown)


bevans2000:
Merged this as the fix was on a separate branch

Issue #527, opened 2021-12-10
Why a big jump in pulse width from 1024x to 2048x  


mokun:
**Is your feature request related to a problem? Please describe.**
- While looking into how best to tweak the value `processInterval` in timePassing() in ResourceProcessing, I came across one strange phenomenon with the pulse width given by `pulse.getElapsed()`.

**Describe the situaion**
- A pulse width of 0.01153474723200986 is ideal.
- At 1024x time ratio, it can be kept at 0.0116.
- However, as soon as player increase the time ratio to x2048, the pulse width would jump up to as much as 3.37932048 and a log statement below appears 
> Proposed pulse 23.34633 clipped to a max of 3.37932048.. 

- This is because of the recent change in simulation.xml as follows: 

```xml
<!-- The time span covered by each simulation pulse is defined below. These are defined in terms of
		     milliSols. The lower the value, the more accurate the simulation but the more pulses & more expensive to calculate. Note both values can be adjusted if the processing becomes too excessive.
-->
<!-- Default: 0.0112644016 millisols ~ 1 earth seconds -->
<min-simulated-pulse value="0.0112644016"/>

<!-- Default: 3.37932048 millisols ~ 5 earth minutes or 300 earth seconds -->
<max-simulated-pulse value="3.37932048"/>
```

**Describe solution you've considered**
- Design the pulse width to increase gradually all the way up to 3.37932048 millisols, instead of making an abrupt change to jump to 3.37932048 

**Additional context**
```
00-Adir-01:180.666 (Config) MasterClock : Time-ratio 512x -> 256x
00-Adir-01:180.666 (Config) MasterClock : The scale factor becomes 28.6
00-Adir-01:185.750 (Info) ResourceProcessing : pulse.getElapsed(): 0.02306949446401972  accumulatedTime: 26.516776967189863  processInterval: 26.512727479497716
00-Adir-01:186.615 (Info) ResourceProcessing : pulse.getElapsed(): 0.02306949446401972  accumulatedTime: 26.525252192444437  processInterval: 26.512727479497716
00-Adir-01:186.661 (Info) ResourceProcessing : pulse.getElapsed(): 0.02306949446401972  accumulatedTime: 26.535783822321182  processInterval: 26.512727479497716
00-Adir-01:187.169 (Info) ResourceProcessing : pulse.getElapsed(): 0.02306949446401972  accumulatedTime: 26.521465367343964  processInterval: 26.512727479497716

[...]

PR #536, opened 2021-12-23
Remove pulseUI from ClockListener

bevans2000:
This PR addresses some performance issues with the pulsing and simplification of the code base.

First change is the pulseUI method is removed from the ```ClockListener``` interface. Logic is now moved to the existing ```clockPulse``` method. For UI listeners the listeners now have the ability to specify the minimum duration between the delivery of pulses which allows these listeners to work at a slower pace to the main clock. Additionally different UIs can have a different pace, e.g. the TimeTool has a high pace.

Second change is to move the Simulation saving logic out of the ```MasterClock`` and have it solely resides within the Simulation class. The clock should just focus on modelling time and notifying listener to changes in time. The autosaving is now archived by having a dedicated ClockListener that has a duration equal to the auto save period.; hence it fires once per X minutes. The saving is now performed at the end of the Simulation.clockPulse method if there is a save pending

As an aside I did get a NPE from deep inside the Swing UI components but it was not possible to track it's origin because it was triggered via invokeLater in Swing. I think it has nothing to do with these changes. I suspect a window has been changed by a pulse but there was a background refresh queue in the AWTThread.
#527


mokun:
By design, if the time ratio we want is way too high for the machine to handle, it would reset the time ratio back to 1x.

This happens more often now since the commit.

At 1024x, it starts to happen sometimes

bevans2000:
I feel the root cause us probably more pulses running but I haven't tried any high ratios yet. I might try after Christmas. Wide pulses are certainly a problem.

mokun:
> Wide pulses are certainly a problem.

The sudden increase of the width of the pulses is one problem.

But if we allow the gradual increase of the width of the pulses, then it's not as bad. 

So, if we can *compute* how bad the system stability has become, we can set a limit to how high the time ratio is allowed for each particular machine.   Can we ?

bevans2000:
Yes so the calculations adapt to the actual system being used. It does that to some extent now by using the accuracy setting but that was when the simulation was a lot slower. We have optimised a lot and the approach needs revisiting (& simplifying)

Issue #280, opened 2020-10-08
Food production values


Paculino:
I started adding onto the food production XML (fixed a typo "slat" to "salt") as well as the resources and meal XMLs, and noticed some discrepancies where the output amount did not seem to match the input. The major discrepancy was for rice flour, which based upon the XML is 80 % water by mass. Should it not be more similar to soy flour in this manner? The other discrepancies made more sense, as water can boil off, or food waste, and about these I am only curious as to whether or not the sim takes these into account when calculating partial pressure of water vapour and the waste production.

mokun:
@Paculino , 

I suppose you're looking the process here in [foodProduction.xml](https://github.com/mars-sim/mars-sim/blob/06f3c3550541e180d291d99b57746d5fbc17fa03/mars-sim-core/src/main/resources/xml/foodProduction.xml#L374-L387) 

Great that you pointed this out we may need to adjust the amount of water. 

So the question is, what the amount of water by percentage of the input weight when producing rice flour ?

> The other discrepancies made more sense, as water can boil off, or food waste, and about these I am only curious as to whether or not the sim takes these into account when calculating partial pressure of water vapour and the waste production.

Yes I assume it will go through some cycles of boiling and a lot of water will be lost. For now, it was just simple input and output.

Q: are you able to find any good reference out there in how much water is needed ? 

I saw this [one](https://scholar.google.com/scholar_lookup?title=Mass+production+method+for+rice+protein+isolate+and+nutritional+evaluation.&author=Morita+T.&publication_year=1993) but it talks about only the **dry mass** and gives the breakdown of protein, carbohydrates, etc. We haven't gotten down to that level of nutrition modeling yet.

> whether or not the sim takes these into account when calculating partial pressure of water vapour and the waste production.

Not yet. We didn't consider the specific type of waste being produced when doing food production. No calculation on how the water vapor may be recaptured.

After each meal is being consumed, the kitchen will create some generic amount of food waste and that's it.  

For sure, we can model better if you can come across any useful reference

Do by all means post any link you want us to use to improve our modeling.

Paculino:
[...]

PR #668, opened 2022-08-17
Soy expansion

Paculino:
Since I was working on finding sources to check the accuracy of my personal xml modifications, I thought that it might be good enough to try to contribute it, so here it is. Soymilk and tofu yields were modified to match the average result from a paper mentioned in a comment in the foodpProduction.xml , and vinegar was added as an alternative coagulant for tofu. Since okara's presence when attempting to coagulate tofu prevents much of the coagulation, leaving more of a gritty porridge, it would be a an unused edible by-product of soymilk and tofu production. Because of this, a meal using it, granola using it, muffins using it, and a process to ferment it (or whole soybeans) into tempeh (which can be used instead of tofu sometimes) were all added. Since honey may not be readily available on Mars (whether it be from bees dying off or simply never bringing bees), sugarcane juice can be substituted. Lastly, miso and ketchup were added, and are used in meals where you would expect them. This partially addresses [issue 280](https://github.com/mars-sim/mars-sim/issues/280) on major discrepancies in mass, but also addresses the comment on scaling up production, since efficiency of soy is increased.

### DIFF
--- a/mars-sim-core/src/main/resources/xml/foodProduction.xml
+++ b/mars-sim-core/src/main/resources/xml/foodProduction.xml
@@ -165,6 +165,111 @@
 		</outputs>
 	</process>
 
+	<process name="Package Quinoa into Granola Bar"
+			tech="1" skill="0"
+			work-time="50"
+			process-time="100"
+			power-required="0.1">
+		<description>Package puffed Quinoa, sugarcane juice, nuts and dried fruits into Granola Bar.</description>
+		<inputs>
+			<resource name="Quinoa" amount="4.0" />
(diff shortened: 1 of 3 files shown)


mokun:
@Paculino ,

Great to hear from you again !

> foodpProduction.xml

Thanks for being detail oriented in soy products

>  Since honey may not be readily available on Mars (whether it be from bees dying off or simply never bringing bees), sugarcane juice can be substituted.

We didn't finish the bee modeling. Will you be interested in working on that ? 

Let's open a new thread for discussing how we may add this feature.

PR #753, opened 2022-11-16
Create FactoryMetaTask class

### LINKED ISSUE #740: TaskMeta.getProbability should return an object instead of a double
Some of the Taskmeta classes calculate target entities as part of the getProbability method. These are the target of the appropriate Task. For example the RepairInsideTaskmeta selects a Malfunction and then pushes it into the Settlement object which is picked up in the RepairInsideTask constructor. Essentially this is caching within the Settlement. This can create problems with cached entities becoming out of date or stale.

A cleaner solution is for the elected Entity to be passed as the result of the probability calculation, a new class containing the probability score and an optional object. This would be saved in the TaskMap and then the object would be passed to the actual Task in the createInstance method.
This removes the need for a Task to recalculate the Entity to execute the Task on.  
The majority of the TaskMeta do not use a caching mechanism and will continue to work based on the probability score.

bevans2000:
This PR creates the ```FactoryMetaTask``` class that represents any MetaTask that can act as a factory to create the associated Task instance. This involves a reparenting a lot of the existing MetaTask to remove the inconsistency in the new approach.

close #740


bevans2000:
@mokun Hang on something has gone wrong here. One of the changes are missing. I'll fix this

bevans2000:
@mokun This is my mistake I missed a file in my last commit by accident and that is why the build failed. I created the pull request without waiting for the build which was my fault.

BUT you should not have been allowed to complete the merge. I've realised that our repo does not have Branch Protection enabled which prevents the merging of failed code.
I've enabled branch protection on the master branch such that any Pull Request must successfully complete the build status check before it can be merged. I thought this check was enabled by default.

PR #760, opened 2022-11-23
Implement flexible Shift Patterns

### LINKED ISSUE #743: Make flexible Shift Patterns
**Describe the bug**
Shifts are not flexible but hard coded within the simulation. There are 2 shift patterns supported (A,B) & (X,Y,Z) where the start times are hard coded. The shift pattern for a Settlement is selected according to the number of settlers. 
The purpose of the Shift patterns is to have a main daily time shift but as Settlements round round the planet this concept breaks down.
Secondly, every person has a ```TaskSchedule``` which contains logic that is more suited to the Shift than the Person. This is further complicated by ```Settlement``` also having similar logic.

An alternative is to remove the hard-coded shift settings and add the concept of shift patterns to the settlement configuration. Shift patterns could be extended to include the concept of repeated "off Shift"/Rest days.
On Call would be a special shift automatically added.
The settlement templates would reference which Shift Pattern to use; this would also accommodate the time zone of the Settlements Locations, e.g. morning sunrise shift is still that on opposite sides of the planet.
To support this would be the creation of some new configuration classes but the big change would be the replacement of the individual ```TaskSchedule``` with a reference to a Settlement ```Shift``` instance used by everyone on the same shift. The Shift object would control when Persons change Shift; not the individual as is the case now. Probably a new ```ShiftManager``` object for the Settlement that also controls who has a day off.
There would be some impact on TaskManager choosing the potential TaskMeta instances.

bevans2000:
This PR allows the shifts in a Settlement to be configured externally as ```ShiftPattern```. These patterns can be shared between ```SettlementTemplate```. They have are used to configure a ```ShiftManager``` that controls shift allocations in a Settlement. Each manager has the following attributes:

- Set of Shifts which are offset according to the SunRise at the location
- Will auto allocate workers to Shift to match the Shift target %age target of the population.
- Ability to select a subset of Workers periodically and move them to a different Shift
- Workers are given a day leave when transitioning between Shifts

There is a new Console command 'shifts' to show the details of a Settlements Shifts.
The changes are tested by a new dedicated UnitTest.
The hardcoded logic has been removed with a configurable based approach. This comes with a number of predefined Shift patterns

- 3 Shift Standard - 3 even shifts covering the full day
- 2 Shift Standard - 2 even shifts covering the full day
- Night Skeleton - 3 Shifts but the night shift is long and has a small percentage target of persons
- 3 Shift Long - The 2 day shift have long durations with an overlap.

Close #743


mokun:
It's good that now it shows at what millisol the work shift end in Person Window's Activity tab.

image

Question: Can we show at what millisol the work shift starts ?

bevans2000:
> Question: Can we show at what millisol the work shift starts ?

Yes, I created a new method in ScheduleTabPanel called createShiftDescription, i think, that creates the string representation for multiple places. Just change it there

Issue #820, opened 2023-03-16
Upgrade to Java 17


bevans2000:
**Is your feature request related to a problem? Please describe.**
Project is currently compiling to Java 11. Java 11 is a LTS and will be actively supported until Sept 2023.
However many in the Java community are using Java 17 (LTS) as a baseline. This release has longer support.

There are many benefits in Java 17 but one key one to this project is the introduction of the [Record keyword](https://www.baeldung.com/java-record-keyword). The code base has many transient POJO which would benefit in being converted to Records to reduce the code base size.

WebLaf was previously seen as a blocked but this has been completed. #789 

**Describe the solution you desire to see**
Upgrade the java version in the Maven pom files.
Change the Docker images to use Java17.

**Describe alternatives you've considered**
Java 21 is the next planned LTS but this has not been released yet.

bevans2000:
I think the 1st step is to review the dependencies in the various pom.xml and remove any unneeded JARs.

bevans2000:
The SonorQube flow is failing. I'll fix it, obviously the default Ubuntu image does not have the latest Java installed.

bevans2000:
Problems with Eclipse IDE

bevans2000:
So Eclipse is running Java17 JVMs in a different way to everything else. Standalone works as does VSCode but Eclipse is tripping up on the module info and imports. The underlying problem is the batik libraries have copied the javax.xml code into their own JAR file which is a big security issue and Eclipse blocks this. I've got round the problem. I have done:

- Rebuilt in a new Eclipse IDE workspace using the internal JVM
- Command line via Maven using the Amazon Corretto OpenKDK
- Clean build and run in VSCode IDE using Corretto
- GitHub maven gitflow using Eclpse Temurin OpenJDK

PR #823, opened 2023-03-23
Convert to Java17

bevans2000:
This PR upgrades the minimum Java to 17.

This involves changing the ```release``` property main POM.xml. To prove the change a few classes have been changed to use Java 17 features.

In addition the Dockerfiles have bee changed to use a AWS Corretto Java17 image. The Docker OpenJDK images are now deprecated.

Note this will require a Java 17 installation to run or build the project.

Part of #820

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -42,7 +42,7 @@ jobs:
 
     # Initializes the CodeQL tools for scanning.
     - name: Initialize CodeQL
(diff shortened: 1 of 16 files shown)


bevans2000:
The CodeQL is failing so I will have to check that tonight. Looks like the version we are running is deprecated but the doesn't explain the failure.

bevans2000:
Got there. I have had to change how the CodeQL flow works.

PR #874, opened 2023-05-11
Maps are downloaded from a remote site

### LINKED ISSUE #867: Separately download and install elevation data and map data type
- Currently, the mars-sim app is bundled with a very low resolution elevation data set. It's only 4 pixel per degree (PPD) in resolution and is 2 MB in size. The next step up is a 16 PPD data set and is 32 MB in size.
> EDIT :
> - From time to time, a rover cannot find a route to go and get stuck in going home or going to a destination. 
> - This happens as the slope angle becomes too high for the vehicle to maneuver or to climb. 
> - This is attributed to the fact that low resolution of elevation data set is being used that can cause an abrupt jump from, say, 0 km to 3 km, in an instant.

**Describe the rationale**
- It provides higher fidelity in modeling and better flexibility to end users when the use of various resolutions of the elevation and map data sets are being offered.
- It will help keep the core mar-sim package lean in size.
- It is more manageable in development when the offering of mars-sim app and extra data sets are separate.

- Prepare a few additional jars with different resolution of elevation data and map data. 
- Create a menu option for managing and choosing a different elevation and map data set.
- When running the core mars-sim program, it will check on the default location in `.mars-sim` to see if at least one an elevation data set and one map data set exist in the local machine. 
- If available, it will proceed to start mars-sim, if not available, it will direct the player to download and install a data set separately.

**Describe alternatives you've considered**
- Bundle a higher resolution elevation and map data sets together with the core mars-sim app. The entire package will no longer be **portable**.
- This would invariably create a barrier in some remote part of the world in downloading a large program and timeout would occurs.

**Specifications**
 - v3.5.0

**Additional context**
- See recent [commit](https://github.com/mars-sim/mars-sim/commit/f73e4f777858a922144ad3c9a89c01bd85690f47) in testing the use of a 16 PPD elevation data set.
> EDIT :
> - Need to look into some scenarios in which the waypoint destination is, say, in the middle of crater, the vehicle may not be able to find a route to go down. 
> - The same can also happen if a settlement happens to be at a very low or high elevation spot. It will take a lot of energy to climb up to a peak or climb out of a valley.

bevans2000:
This PR allows map images file to be downloaded from the internet and cached to the local file system.

Map types are now totally data driven by the types defined in the ```mapdata.propeties```. The using logic in the Map classes are driven by the user's selection with no hardcoded logic for specific map types any more. This revolves around a new ```MapMetaData``` class that describes the attributes of a map type including a new ```isColourful``` attribute that influences the styling of icons and labels added to the map.

The user is asked to confirm the downloading of a new map type if it is not cached; this handles a slow download.

fix #867


bevans2000:
The Unittest fails because the MEDG image file is not on the master branch yet to trigger the download during the testing.

Issue #919, opened 2023-06-16
Task difficulty Level - Ask for help when performing a difficult task


mokun:
**Is your feature request related to a problem? Please describe.**
- A difficult task should trigger the need to request help.
- Simulate more relational aspect sim by allowing tasks to be done with more interactions between settlers. 

**Describe the solution you desire to see**
- Allow provision for a person to request help from another person when performing a task.
- This can be just a remote call to talk to someone or asking that person to come over and offer physical assistance at a location.

**Describe the rationale**
- Each particular task in a situation should be coded with a particular difficulty level.
- It's not natural that almost all the task in mars-sim involve a single person when performing a task, 

**Specifications  (please complete the following information):**
 - pre 3.6.0 build 8187

**Additional context**
- Will need to come up with a mechanism to quantify the difficulty of a task.
- If the task is more difficult than the skill level of a person, he's more than likely to request help. 
- When help is received, relationship score should also be adjusted.
- Relationship score should slowly decay over time especially when there is no interactions between two person.

mokun:
Related to https://github.com/mars-sim/mars-sim/issues/1111.

PR #920, opened 2023-06-16
Delivery a new approach to managing Missions

### LINKED ISSUE #890: Create a new TestDrive Mission
Create a new Mission instance based on the new Mission Project approach. This will cover the basic steps of a Mission:

1. Review
2. Load Vehicle
3. Embark
4. Drive 
5. Disembark

bevans2000:
This PR deliveries a new approach to managing Mission. 

This new approach is demonstrated in a new Mission called ```TestDriveMission```. This mission can only be created via the console _create mission_ command at the moment.
The new mission type is fully integrated into the Monitor & Mission tools and the console commands

The new approach concept represents the Mission as a number of seperate ```MissionStep``` where the new ```Project``` class is used to progress and advance throught the steps.
This provides a very easy to understand and flexible. There are a number of common shared steps:
* ```MissionTravelStep```  - involves driving from one ```NavPoint``` to another.
* ```MissionBoardVehicleStep``` - handles getting members to board a vehicle
* ```MissionLoadVehicleStep``` - handles loading a vehicle, could be the start or in the middle as in a Trade mission.

To accompany the changes there calculation of requires "supplies" has been changed. Each unprocessed step is queries for the suplies it needs. These are captured in a new ```MissionManifest``` class.

Note the current implemented steps do not have the full feature of the old Missions, e.g. no handling of Vehicle malfunctions.

Close #890


mokun:
Finally the Project Mission delivered !

bevans2000:
Yes took a while but I'm very happy with the outcome. The nature if using a set of steps is really clean

mokun:
Yes I can appreciate that there's no more crazy calling of
determineNewPhase() and performPhase() !

bevans2000:
I want to convert the dig missions next as that will introduce an sharable EVA step. 
Also need to do

- Checking resources during the mission
- Selecting vehicles
- selecting members

mokun:
> dig missions next as that will introduce an sharable EVA step.

I like that. Will there be a division of labor in these types of outside EVA activities ? 

I suppose in terms of building a weather station, there will be many activities involved and person A may be doing X and Y, while person B may be doing Y and Z and person C may be doing X and Z, etc.

That's why if we can implement the idea of the **shared** task first (please see and comment on #919), then it may get really easy to implement capture the spirit of a sharable EVA step.

What would you say ?

mokun:
@bevans2000

Do you want to continue on this thread ?

bevans2000:
This pr is closed now

Issue #650, opened 2022-07-30
Modelling of time


bevans2000:
**Describe the situation**
Earth time is modelled using a custom ```EarthClock``` class which has numerous bespoke methods. Java has an off the shlef way to model earth time using the existing time package, e.g. LocalDateTime. We should totally remove EarthClock and replace it with the standard Java time classes.

Secondly, ```MarsClock``` represents an every moving time value. Time should be modelled as a fixed immutable object representing a fixed moment in time, as in the LocalDateTime implementation. We should replace the clock approach with a MarsDateTime which is equivalent to the standard earth date time. This would remove the need to call the ``clone``` method excessively and guarantee that there is only one representation in memory for the same point in time.

The MasterClock would be responsible for providing the current Earth & Mars time going forward. This would be automatically updated in each clock pulse.

mokun:
> Secondly, `MarsClock` represents an every moving time value. Time should be modelled as a fixed immutable object representing a fixed moment in time

Yes. Currently we have two constructors in MarsClock. This can be confusing. One has moving time value and has been used in too many class/methods. The other does have one fixed time value only stored inside.

mokun:
> Java has an off the shlef way to model earth time using the existing time package, e.g. LocalDateTime.

How do we use Java's time package to construct a new MarsClock ? 

Note that the timekeeping methods on Mars are very different from Earth and there's no standard Martian time yet.

We are defining our own in-house standard things as we speak.

Also, we haven't considered how to implement Time Zone in mars-sim yet. 

Since there's no national order on Mars yet, we could easily divide Mars into 12 time zones based on longitude, as some has suggested.

bevans2000:
We only use the standard Java time packages to replace EarthClock. We build our own to represent Mars time

mokun:
Do you still like the Earth time and calendar, namely, `earthDateField`, on the top right of the Main Window ?

It requires Weblaf's component `WedDateField` but their development is very slowly. @Urwumpe suggested in the past that we switch UI layout manager to something else but they don't quite support Java 17 and uses reflection.

Any suggestion to rebuild that Earth time and calendar with something similar or better ?

bevans2000:
I think having just the Mars time is sufficient.

Urwumpe:
I think we only need to support Mars time for more interaction, and maybe have GMT time as simple text field somewhere for reference. A custom Mars date picker or time-picker would be nice to have, but not high priority, since we rarely need to select a time manually.

PR #948, opened 2023-06-24
Update Meta mission

bevans2000:
This PR updates the MetaMission interface to remove the static references. It's generally not  good idea to have static references to mangers inside an interface.
The functionality has been relocated to the AbstractMetaMission with the addition of some consolidation of duplicated logic.
Also it has been converted to the immutable MarsTime class.

Part of #650

### DIFF
--- a/mars-sim-core/src/main/java/org/mars_sim/msp/core/Simulation.java
+++ b/mars-sim-core/src/main/java/org/mars_sim/msp/core/Simulation.java
@@ -583,8 +583,6 @@ public class Simulation implements ClockListener, Serializable {
 		CropConfig cc = simulationConfig.getCropConfiguration();
 		MedicalConfig mc = simulationConfig.getMedicalConfiguration();
 		
-		// Gets the MasterClock instance
-//		MasterClock masterClock = masterClock;
 
 		// Gets the MarsClock instance
 		MarsClock marsClock = masterClock.getMarsClock();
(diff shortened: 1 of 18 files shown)


mokun:
> generally not good idea to have static references to mangers inside an
interface.
Point taken !

Btw, I've found out Regolith mission doesn't work. Can you take a look at
CollectRegolithMeta.java
<https://github.com/mars-sim/mars-sim/pull/948/files#diff-1fbcc0e5b252029280b2c6d9680e0d36f10d68c8b5cb0e92c21195ff2e9b53af>
since
it fails to scoop up any regoliths at all.

bevans2000:
Does the regolith mission start but get aborted because it can not find a site or does it never get started?

mokun:
> Does the regolith mission start but get aborted 

Yes they can start. 

They never abort.

> because it can not find a site or does it never get started?

They did find all the sites.

However, they never retrieve any regolith and they pack up and go home.

bevans2000:
Ok. That gives me a start on where to look

bevans2000:
I can't find anything obviously wrong; the missions are failing because there is a lack of members that can be recruited.
However I have noticed that there is a massive imbalance between the Exploration & Regolith/Ice missions; the later have a very small probability, less than 1 but the Exploration has probability over 20. This means there is a lack of resource collection missions being started. 
I have found out why this is happening.

bevans2000:
It's because the GoodValue for mineral is returning a high value which is giving the Exploration mission a high probability. I've adjusted the values but obviously this is very sensitive to changes to the Goods calculates.

Issue #947, opened 2023-06-24
Restructure of Mapping classes (into `mars-sim-mapdata`)


bdowrick:
**Is your feature request related to a problem? Please describe.**

Currently in the `mars-sim` project, a number of map features are within the `mars-sim-core` (or worse, the `mars-sim-ui`) module and allow for usage of such things as `Coordinates`, `LocalPosition`, `LocalAreaUtil`, `OrbitInfo` for location and information.

Some features of `mars-sim-mapdata` are duplicated over into `mars-sim-ui` as well, such as `Map` vs `MapData` which are near identical in usage, with minor changes between them.

**Describe the solution you desire to see**

A complete restructure of _all_ geo/positional related tools and classes into `mars-sim-mapdata` and create util/tool classes for `mars-sim-core` to extend and populate the world with data.

_Side note:_
There are some shared classes that would also need to be moved over into an additional module `mars-sim-tools`

**Describe the rationale**

While this is a considerable change, and may require a fair bit of work to actually implement, I believe it's the best way to move forward with `mars-sim` as it allow for reduced code duplication, lets the `mars-sim-core` focus solely on the model of the simulation, and `mars-sim-ui` focus on UI.

**Screenshots**

![image](https://github.com/mars-sim/mars-sim/assets/16632004/ab9fb032-e5ae-46c2-a68b-d3a96fc4fac4)

bdowrick:
@mokun @bevans2000 

Working through #935 has lead me into thinking this may be a better way of doing things, as more than just `MineralMapLayer` would benefit. Allowing me to improve performance of all layers (and anything location related) a lot easier.

(Sorry for delay as well, I've been sick past few days)

bevans2000:
Totally agree. I have been thinking something similar for a while..
The logging classes and directory management should go into this tools package.
I think the change isn't too bad, obviously it needs to be done on a branch

mokun:
Hope you feel better with good reat.

Yes let's create mars-sim-tool submodule and  separate concerns better and make things cleaner.

mokun:
Good that `mars-sim-tools` is created.

![image](https://github.com/mars-sim/mars-sim/assets/1168584/21f41b53-1234-4806-aa75-796ef7dc6b06)

So far, there are only two class files in it.

Any other class going there as well ?

bdowrick:
I plan on moving OpenCL to it, alongside any other classes that may be useful to other modules.

[...]

PR #952, opened 2023-06-26
#947 - Mapdata restructure & mars-sim-tools creation

bdowrick:
This is an intial pull request to move the structure of certain classes around and add the modules!

Package structure is terrible currently, but to change this before the modules exist makes it a lot more difficult to manage.

_Also fixed a bug in NavigatorWindow.java where the unitManager is set to null before a function is called_


mars-sim-core/src/main/java/org/mars_sim/msp/core/tool/SimulationConstants.java
```diff
@@ -0,0 +1,15 @@
+package org.mars_sim.msp.core.tool;
+
+import org.mars_sim.msp.core.Simulation;
+import org.mars_sim.msp.core.UnitManager;
+import org.mars_sim.msp.core.environment.SurfaceFeatures;
+import org.mars_sim.msp.core.person.ai.mission.MissionManager;
+
+public interface SimulationConstants {
+
+  static Simulation sim = Simulation.instance();
```
bevans2000, line 10: In the future I want to reduce the dependency on ```Simulation.instance``` as it can cause problem when a simulation is loaded.
The simulation code generally relies too much on static references but there is no need to change it yet.

Issue #958, opened 2023-07-01
Members never bring their EVA suits if boarding a mission vehicle in garage 


mokun:
**Describe the bug**
- For Alpha Base's missions, none of the occupants have EVA suits in the rover Dusty Star. They can't go out to do regolith prospecting.

**Expected and unexpected behavior**
- Case 1: No garage in a settlement :
When boarding the vehicle, each member already has already donned an EVA suit. They will doff the suit in the vehicle. This isn't the issue.
- Case 2: Have garage and board the vehicle parked in garage :
Need to ensure `RoverMission` manually transfer sufficient # of EVA Suits from settlement to vehicle. This is what's missing.

**Java classes involved**
- RoverMission, ExitAirlock, Walk

**To Reproduce (Please provide steps)**
- Steps to reproduce the behavior:
1. Observe how the Alpha Base does a mission and check if there are any EVA suits stored in a vehicle during the trip.

**Stacktrace**
```
01-Adir-03:053.576 (Warning) EVASuitUtil : [0.0272 N 0.0618 E - Dusty Star] Hanna Szabó - Could not find any EVA suits in Dusty Star.
01-Adir-03:053.576 (Warning) ExitAirlock : [0.0272 N 0.0618 E - Dusty Star] Hanna Szabó - Could not find a working EVA suit and needed to wait.
01-Adir-03:056.909 (Warning) EVASuitUtil [x5] : [0.0272 N 0.0618 E - Dusty Star] Emma Pichler - Could not find any EVA suits in Dusty Star.
```

**Specifications (please complete the following information)**
- pre 3.6.0 build 8335

**Additional context**
- One could use enteringRoverInsideGaragePhase() in Walk to transfer EVA suits from settlement to vehicle prior to departure.

bevans2000:
The correct fix is to the EVAsuits the the required equipment list for the Mission if the vehicle is not in a Garage

mokun:
> The correct fix is to the EVAsuits the the required equipment list

Good point.

We could do both. 

The required equipment should have EVA suits

Then after loading, if it's still missing during departure. I don't mind doubling check it.

Also, what if the settlement doesn't have enough (i.e. not everyone have an EVA suit). 

We'll need to decide the ratio of # of EVA suits in a vehicle versus the# of EVA suits staying in a settlement.

mokun:
I just found this snippet in `RoverMission` : 

```java
	/**
	 * Performs the departing from settlement phase of the mission.
	 *
[...]

Issue #888, opened 2023-05-21
Create a standard approach to hold values that change over time


bevans2000:
**Is your feature request related to a problem? Please describe.**
The simulation currently hold a number of properties that change over time where the history is needed to be recorded, e.g. Mission log, Vehicle Status, Task schedule.
Each one uses a different mechanism and needs direct access to a ```MarsClock```. 
Other properties may benefit from the ability to record the history of changes.

**Describe the solution you desire to see**
Create a standard class using Generics that can record changes to a value over time. It would be related to the existing ```SolDataList``` logic. The class would record changes over time.
It would automatically tag the entries with a timestamp when the are inserted. In addition a fast access method to the last value would be provided.
This new class would be retrospectively fitted to the historical properties.

bevans2000:
Related go #650

mokun:
> It would automatically tag the entries with a timestamp when the are inserted. In addition a fast access method to the last value would be provided.

So in #917 we mentioned we could make MarsClock immutable. 

But if that's the case, will it still an advantage to those history logs that use MarsClock when we can't let these MarsClock instances gabage collected ?

Should we rather save those logs in JSON as separate text files and make these JSON log uniquely associated with a certain sim session ?

bevans2000:
It makes no difference if we save in JSON or not. This is about how we model the values in memory. JSON is about how we persistent them.
Garbage collection is better when using MarsTime because a point in time is only represented by a single immutable object in memory, normal good design. Currently we create multiple instances of MarsClock that represent the same point in time.
Anyway I have the new MarsTime working on a branch where it works in parallel to MarsClock

bevans2000:
This can also cover Role assignment #921

bevans2000:
@mokun I have been looking at the  various way we store changes over time and the majority we are storing history which is optimised to ui at the expense of effecient memory usage. However the vast majority of these histories are never viewed in a ui so the extra storage is wasted.

So what I would like to do in consolidate it to use the new ```History``` class in most cases. But create a dedicated ui panel to show any History object. This would include centralizing the filtering/selection into the panel. This would help our UI code 

I think we could use at least in the following areas:

- person job and roles in the career panel
- vehicle status log
- mission log
- person activity log

mokun:
> However the vast majority of these histories are never viewed in a ui so the extra storage is wasted.

Yes. It can be a waste if we don't document what those logs are and how to use them to the advantage of players.

[...]

PR #960, opened 2023-07-02
EVA Suit has history

bevans2000:
This PR add a History to the EVA Suit so that the location is recorded. The history is shown in the EVA general tab.

There is a general tidy up of the ``setContainerUnit``` in the Unit class. Only units that can move need to support the changing of a container and hence the set method has been removed from the base Unit class.

In particular, for Equipment the method was called randomly but should only be used from the transfer method. This may account for the leakage of EVASuits,

The selection of EVASuits for Vehicle & settlements use the same logic; the previous Vehicle logic would never select an empty suit as the resources feel below the mission threshold. 

Due to the important of the EVASuit fix I haven't done as much testing as normal for such a large PR. Hence there may still be some issues.
Relates to #888 & #958

### DIFF
--- a/mars-sim-core/src/main/java/org/mars_sim/msp/core/Simulation.java
+++ b/mars-sim-core/src/main/java/org/mars_sim/msp/core/Simulation.java
@@ -349,12 +349,12 @@ public class Simulation implements ClockListener, Serializable {
 		unitManager.addUnit(outerSpace);
 		
 		// Initialize Moon instance
-		Moon moon = new Moon();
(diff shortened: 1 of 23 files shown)


mokun:
May be I'll add another column called *Locale* to show what the top container is. 

If it's on a person, I'd like to resolve if a person is in a building in a settlement, on a vehicle inside a garage or outside on Mars surface, etc.

bevans2000:
I don't think its needed really

PR #966, opened 2023-07-08
Fixed typo in TabPanelGeneral.java

Paculino:
Human height as listed in the general tab of the details window is shown in cm but labelled as m; I changed the label to cm to prevent the >100 m tall giants.


bevans2000:
@Paculino thanks for the fix...... although I lime the idea if 100m tall giants on Mars 😆

bevans2000:
@Paculino feel free to pick up any of the issues or purpose changes/improvements

PR #1033, opened 2023-09-04
Fix finding an empty activity spot. Moderate tending crop/fish  

mokun:
1. Moderate tending crop and fish 
2. Correct a major issue of finding an empty activity spot


mokun:
@bevans2000 

I never wanted `Merge branch 'master' into 3.6.X`. 

Can you reverse it ?

PR #1038, opened 2023-09-05
Process Override controlled by Settlement Preferences

### LINKED ISSUE #963: Change SettlementPreferences logic to be more generic to cover Policies
The modifiers used to score various activities in the simulation in terms of a Settlement are controlled by a common set of Preferences. These can be modified/viewed from the UI.
There is also a parallel set of controls for Process Override that have separate logic.

The Override can be transformed into a modified version of the Preference logic. This improved logic could define a set of Policies for a Settlement that could be either a flag or a modifier. The benefit is this would be a single solution to cover all cases and be very generic.
The Type field in the PreferenceKey class would define if a double or boolean value should be held; hence influencing the UI.
This would deliver a generic solution that can be used elsewhere.

bevans2000:
This PR expands the Settlement Preference logic to replace the specialised Process Override logic. The latter is now a new ```PreferenceCategory`` and hence editable from the Settlement Preference panel.

The Preference now supports boolean values as well as the existing double. The preference value type is defined by the PreferenceCategory such that all preferences of the same category have the same type.

In addition, a simplification of the labelling for MissionType has been applied that relies on the enum name  to lookup the name in the message properties.

close #963


bevans2000:
Great

PR #1058, opened 2023-09-15
Rework removing rover from garage in transfer 

mokun:


mars-sim-core/src/main/java/org/mars_sim/msp/core/environment/SurfaceFeatures.java
```diff
@@ -132,15 +133,15 @@ public double getOpticalDepth(Coordinates location) {
 		if (value != null)
 			return value.doubleValue();
 
-//		opticalDepthLock.lock();
+		opticalDepthLock.lock();
```
bevans2000, line 136: It our code based this provides no benefit over a standard (& simpler) synchronized block. ReentractLock is for a specific problems cases and we are just guarding against concurrent modification

Issue #903, opened 2023-06-05
Put settlements/stations/outposts of same space agency under one belt


mokun:
**Is your feature request related to a problem? Please describe.**
- Both Europa and Arcadia are ESA settlement in scenario_default.xml. It would be silly to separate them into different settlement panel in Monitor Tool.

**Describe the solution you desire to see**
- List both Europa and Arcadia under one belt.

**Describe the rationale**
- if a settlement creates numerous weather/communication stations, mining stations, trading outposts, etc. it's more intuitive to find and see the general statistics of them in single location.
- In future, we may even create a dashboard showcasing combined statistics of related settlements/stations/outposts.

**Specifications  (please complete the following information):**
 - pre 3.6.0 build 8131

**Additional context**
- One unintentional effect was found in that two unrelated settlements were put under one panel in #900. We want it to happen under a different circumstance/reason.

mokun:
This design change would be nice in anticipation of having #907 in future.

mokun:
@bevans2000 

Other than ESA, We also have 2 settlements under Mars Society now.

Can you change it in such a way to make the Settlement tab show 2 settlements, instead of one ?

bevans2000:
Do you mean in the Monitor Window ?

mokun:
> Do you mean in the Monitor Window ?

Yes. In Monitor Tool's Settlement tab.

Actually, the original purpose of this thread is to get weather stations, depots, outposts under the jurisdiction of a settlement all under one settlement tab.

I suppose each settlement should have its own settlement Tab.

But now, in view of this, we could build a combo box having a list of sponsors, in addition to the settlement combo box.

Player will be given the choice to select either one of the combobox.

If a sponsor is chosen in the sponsor combobox, then a sponsor tab will be created.

If a sponsor have two settlement, then the two settlement will both be shown in this tab and visual comparison can be made relatively easily in terms of resources, power, etc.

The vec'hicle tab will show all vehicles from both settlements under the selected sponsor.

Do you like that ?

bevans2000:
We could do. It's a lot of work. Is not too tricky for tabs showing Units but harder for the tabs that only show details of a single Settlement like Crops or Goods. 
Secondly we should make a decision on the logn term layout of the UI before making large changes.

mokun:
@bevans2000 

I just created [#1095](https://github.com/mars-sim/mars-sim/issues/1095).

How do we create a combobox that will take in either a list of settlements or a list of sponsors ?

[...]

PR #1074, opened 2023-09-20
Monitor model improvements

bevans2000:
Objective of this PR is to addresses performance and code smells.

- Remove a lot of duplicate code by creating a ```AbstractMonitorModel``` class to hold common methods
- Create a ```ColumnSpec``` record that simplifies the definition of columns by tying name & class.
- In ```VehicleTableModel``` and ``SettlementTableModel``` change how resource driven column and updated/extracted. These now use a mapping from the resource to the column index

### DIFF
--- a/mars-sim-ui/src/main/java/org/mars_sim/msp/ui/swing/tool/monitor/AbstractMonitorModel.java
+++ b/mars-sim-ui/src/main/java/org/mars_sim/msp/ui/swing/tool/monitor/AbstractMonitorModel.java
@@ -1,0 +1,111 @@
+/*
+ * Mars Simulation Project
+ * AbstractMonitorModel.java
+ * @date 2023-09-18
+ * @author Barry Evans
+ */
+package org.mars_sim.msp.ui.swing.tool.monitor;
+
+import java.io.Serializable;
+
+import javax.swing.table.AbstractTableModel;
+
+import org.mars.sim.tools.Msg;
+
+/**
+ * Default implementation of the MonitorModel
+ */
+public abstract class AbstractMonitorModel extends AbstractTableModel
+        implements MonitorModel {
+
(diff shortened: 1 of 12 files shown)


mars-sim-ui/src/main/java/org/mars_sim/msp/ui/swing/tool/monitor/AbstractMonitorModel.java
```diff
@@ -0,0 +1,111 @@
...
+/**
+ * Default implementation of the MonitorModel
+ */
+public abstract class AbstractMonitorModel extends AbstractTableModel
+        implements MonitorModel {
+
+    /**
+     * Helper class to define the specification of a column
+     */
+    protected record ColumnSpec (String name, Class<?> type) implements Serializable {};
```
mokun, line 24: awesome that you are using `record` now.

mokun:
One question is, how do we make Monitor Tool compatible with our future plan of adding a sponsor combobox ? 

See [this]( https://github.com/mars-sim/mars-sim/issues/903).

mars-sim-ui/src/main/java/org/mars_sim/msp/ui/swing/tool/monitor/AbstractMonitorModel.java
bevans2000: I have started using them elsewhere as well. An advantage of upgrading the Java version

bevans2000:
It can be done. Needs some planning and consideration but the MonitorModel will filter of a Set of Settlements instead of one. 
All models then need to be clever to automatically add a Settlement column if more than one is selected.
But I still think the direction of the ui in terms of layout needs considering, #979

PR #1078, opened 2023-09-22
Clean up MonitorModel part 2

bevans2000:
Goal is to clean-up code and remove duplicates

Changes:
- MissionTableModel, BacklogModel & EventTableModel now use the new AbstractMonitorModel
- Allow user to sort on any monitor table column; all column types can be correctly ordered now since the delivery of MarsTime
- Optimise MissionTableModel to effeciently update multiple columns in response to Unit event
- Fix problem in BarChartTab & PieChartTab so it handles values correctly that are null
- Cleaned up BarChartTab & PieChartTab
- Fix sorting problem on BacklogModel due to incorrect column types confusing sorter


mokun:
I was going to say the UI for using BarChartTab & PieChartTab are still not as intuitive and helpful.

When we rebuild our UI in JavaFX or Web, how should we do it differently ?

PR #1080, opened 2023-09-25
Merge SettlementTask, RatingScore and monitor model change from Master

mokun:


bevans2000:
You don't need to create a pull request to update your own branch as its not protected. Just do the merge in git. 
But better practice is to create a new branch when you need it instead of having a long lived one for development

Issue #1084, opened 2023-09-26
Convert Person assignment MetaTasks to Ratings


bevans2000:
Convert the individual ```MetaTask``` implementations for Persons to use ```RatingScore``` instead of a simple double.
This will involve using the new common assessPersonSuitability method

mokun:
So does it mean that meta missions will also be converted ?

bevans2000:
Meta Missions have already been converted.

mokun:
So how do I now convert a regular meta task to the new rating score system ?

bevans2000:
> So how do I now convert a regular meta task to the new rating score system ?

I have a set of changes on a branch which laid down a blueprints of how the new FactoryMetaTask  can be done. I will be creating a pull request soon.

mokun:
So, initially how many meta tasks will have rating score shown ?

bevans2000:
> So, initially how many meta tasks will have rating score shown ?

I think about half at the moment. But the ay I have done this change is ALL MetaTask will create a RatingScore; it's just some will not have a decomposed ratings, just a single base.

bevans2000:
> > So how do I now convert a regular meta task to the new rating score system ?
> 
> I have a set of changes on a branch which laid down a blueprints of how the new FactoryMetaTask can be done. I will be creating a pull request soon.

I was going to convert more but I will work on completing the branch as a blueprint. I've noticed that some of the MetaTask implementations do not seem very scientific when creating a score. As part of this issue we can revisit them.

mokun:
> I've noticed that some of the MetaTask implementations do not seem very scientific when creating a score

not scientific ? Which one ?

bevans2000:
To convert ```MetaTask``` to the Rating Score blueprint the follow changes should be done:

1. Change the signature of the ```getProbability``` method to the ```getTaskJobs``` method. 
2. If no task is suitable return ```EMPTYTASK_LIST```
3. Replace any of the get modifiers with more suitable 'assess*' methods. These methods build up the composition Rating Score.
4. The final return can provide a list of TaskJobs if that is suitable or use the ```createTaskJobs``` method which takes a RatingScore and converts it into a list of  single TaskJob.

bevans2000:
@mokun I've decide to raise the Pull Request now; I was going to do some more testing but I think it might be better to give the changes some different usage.

bevans2000:
[...]

PR #1108, opened 2023-10-26
Person FatorymetaTask Blueprint

bevans2000:
This pull request covers a number of changes with the single objective of giving a user more understanding to how Tasks are automatically selected for Persons.

This converts a number of the ```MetaTask``` instances to use the common 'assess*' method provided to be commonly re-used. MetaTask following the blueprint implement the ```getTaskJobs``` method.

Within ```PhysicalCondition``` the muscle skeleton array is converted to explicit fields to improve readabilitiy.
 
In addition the Activity tab in the Person & Robot unit window has been redesigned so that it can show the alternative Tasks not chosen. Each one has a score where the tooltip shows the score composition.

Part of #1084


mokun:
image

I can see how the composite score of 7.707 came to be for Inam Almutairi. 

The base score was adjusted by 5 modifiers.

mokun:
Interesting that Maria was `Writing Study Collaboration Invitation` and it has 8 modifiers.
image

In that Person Window below, I can see the composite score of 146.36.
image

Can we still have the subtask and show it in TabPanelActivity ?

mokun:
Never mind. I can see the subtask(s) now below : 

image

bevans2000:
> Never mind. I can see the subtask(s) now below :

Yes I collapsed the subtasks to make space for the task choice table. I think it works well

bevans2000:
> 
> I can see how the composite score of 7.707 came to be for Inam Almutairi.
> 
> The base score was adjusted by 5 modifiers.

Yes it's really useful. By using a set of common assess methods the modifiers are consistent and applied in the same way. Do you see that modifiers have a 'x'? Some Tasks have a composite base score as well

PR #1138, opened 2023-11-18
Rework how scheduled Tasks for Persons are managed

### LINKED ISSUE #1029: Scheduled Pending Tasks redesign
Discussed in #1028

Replace the ```ScheduleManager``` with the ```ScheduledEventManager``` instance on the Settlement.  There will be a new ```ScheduledEvent``` implementer that will take the name of the required Task and convert it to a pending Task when the time arrives.
Then rework the ```Settlement.setAppoinmentTasks``` to create a Rota when a Shift comes online; at that point schedule the planned activities. 

There is still a problem with the pending Task "hiding" any urgent Tasks that are awaiting action, e.g. Malfunction Repair but that is a different issue.

bevans2000:
The key change is to remove the Scheduled Manager of the Person and use the existing Future Schedule Event pattern of the Settlement. This requires the addition of a new class ```PersonFutureTask``` that is a schedule event that will create a new Task when it fires.

In addition, the execution logic of Pending Tasks is improved and simplified. The logic is moved to the main TaskManager with the removal of checks based on Task name.

Close  #1029


bevans2000:
@mokun After this commit you can use the Console 'future' command to see how many Persons have scheduled future tasks. It seems too many to be.

mokun:
> @mokun After this commit you can use the Console 'future' command to see how many Persons have scheduled future tasks. It seems too many to be.

But where can I find the `future` command ?

mokun:
> > @mokun After this commit you can use the Console 'future' command to see how many Persons have scheduled future tasks. It seems too many to be.
> 
> But where can I find the `future` command ?

ok. I can see it now. It requires I connect to a settlement first. Then do `/fe`

PR #1139, opened 2023-11-19
Allow tables in the Console to wrap cells contents

bevans2000:
There is now appendTableHeading method that takes a boolean to allow cell contents to be wrapped on word and punctuation boundaries.

Changes:
- StructuredResponse has new table heading method allowing cell wrapping
- New private method to create a single table row
- Backlog & work commands wrap cell contents

@mokun I have implemented wrapping of the cell contents in the Console. You can add it to selected tables now.

### DIFF
--- a/mars-sim-console/src/main/java/com/mars_sim/console/chat/simcommand/StructuredResponse.java
+++ b/mars-sim-console/src/main/java/com/mars_sim/console/chat/simcommand/StructuredResponse.java
@@ -9,6 +9,7 @@ package com.mars_sim.console.chat.simcommand;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.regex.Pattern;
 
 import org.apache.commons.lang3.StringUtils;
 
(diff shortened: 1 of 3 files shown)


mokun:
One related question is, how do we make the console window to be resizable ?

mokun:
After my last commit a minute ago, you can see that even though the JFrame is resizable, the black beryl console content window is not. 

Also, I just discover *Spinning Wait* symbol is at the background. It's not needed and is wasting cpu cycles.  

image

mokun:
After I remove the `WaitLayerUIPanel`, the frame resize works now.

image

image

mokun:
One other issue is that if I hit spacebar on a prompt, it's an invalid input.

But it should return an update timestamp. However, it doesn't.

Only when I type in a correct command would it updates the timestamp.

image

Q: what do we need to change in order to fix updating the timestamp on the prompt, whether, we input a valid or invalid command ?

mokun:
Note that I realize the `TopLevel` class's *getPrompt()* method is the one that construct the console prompt. I've just changed that timestamp format last week. 

Also, the `Conversation` class's *interact()* method is the one that calls up *getPrompt()*.

bevans2000:
So good that it resized but we need to do some things to get the tables resizing:

1. Make the StructuredResponse be aware of the console width.
2. The table heading logic can cal late the width of none fixed width columns 
3. Change the table heading definitions in the console commands so they are not all fixed width column

Do you want to create a new Issue to make tables aware of the console width?

Issue #1166, opened 2023-12-05
Reserve an Activity Spot; Guest Beds


mokun:
**Is your feature request related to a problem? Please describe.**
- Currently, it's possible for a collision to occur when two persons/robots are fighting for the same activity spot to go to. 
- For instance, person A identifies a far-away activity spot in another building to go to, it would get started going in the same frame. But since walking may take more than a frame, this worker may still be on the way by the end of this frame.
- However, person B is also interested in going to the same activity spot. If this spot is in the same building he's at, it would take almost zero time for him to get there and claim the activity spot.
- As a result, when person A gets there, he finds the activity spot has already been taken.
- Both person A and person B be shown on the Settlement Map at the same spot.
- However, person B is the one who actually possesses the *title* of this activity spot.

**Describe the rationale**
- Avoid the possibility of contention 

**Describe the goal(s) you have in mind**
- A reservation system should solve this collision problem and ensure only one worker can occupy an activity spot. 

**Describe the solution you desire to see**
- Much like reserving a vehicle, an activity spot should be reserved ahead of the arrival of a worker to claim the spot, just in case that it takes more than a frame to walk there.
- When the spot is claimed, its reservation is released.
- When a spot is released, its reservation is open.

**Specifications**
- pre 3.7.0 build 9052 

**Additional context**
- This issue is in conjunction with the change implemented in https://github.com/mars-sim/mars-sim/issues/1143

bevans2000:
Yes I've been thinking about Bed and how they could use Activity Spots. They need to register who is physically there but also they can be reserved before hand. 
Not sure yet it could be used for Vehicle but it is possible. They question is when? Now or a later release. I think we need to use Activity Spot everywhere, e.g beds etc. And use a single coordinate frame

mokun:
>  thinking about Bed and how they could use Activity Spots. They need to register who is physically there but also they can be reserved before hand.

I suppose both registration and reservation share some similarity but they are also different.

While we want a person permanently registering a bed and set aside for his own use,  we want a person to temporarily reserve an activity spot so that he is permitted to go there.

bevans2000:
Yes if we allow reservations that can be prevented. Beds work the same but and not unreserved when a person leaves. So reservations can be permanent or temporary.
This is all achievable and would simply a lot of the logic by standardising.

mokun:
> lets stop changing Activity Spot in the master branch and create a branch to make these changes.

[...]

PR #1170, opened 2023-12-10
Rework Settlement Map layers

bevans2000:
This aims to rework the map layers in the settlement to clean up the code and make it optimal. The map displays the "items" and then the LabelMapLayer cycle back over the units to render the labels. This means a double scan of all the itmes being displayed. The aim is to display labels as part of the item rendering.

Changes:
- Create ColorChoice which holds a 2 colour pairing used for rendering text and highlights. Removes the need of r2 variables being passed.
- AbstractMapLayer is new and provides reusable label drawing methods including multi-line for long labels.
- Font and ColorChoice is used as parameters for label drawing
- New WorkerMapLayer renders Worker objects according to user selection. Has a callback for ColorChoice to use.
- Selected Worker is always shown with labels.
- PersonMapLayer & RobotMapLayer use WorkerMapLayer to greatly simplify their implementation
- "unit" layers now draw labels as part of the main layer drawing instead of a seperate layer
- Correct problem with labels being auto-enabled when a Robot/Person is selected. Now respects users choice.
- StructureMapLayer & VehicleMapLayer use LocalBoundingObject for placement instead of 6 seperate parameters.
- Numerous code smells corrections
- Delete LabelMapLayer

Helps to support future change for #1166


mokun:
>  the colours are not ideal for Vehicles

True

Also, how do we solve a person's label got overlapped by another ?

Currently it's very difficult to read a person/robot's label when another person/robot is nearby.

How do we cleanly separate one form another ?

Shall we make a label non-transparent ?

This way, they will be on top of one another. Only the selected one becomes  on top and is cleanly viewable.

bevans2000:
Don't know about the overlapping labels but after this change it is clear to see what is going on.

bevans2000:
Also, I now realise I should have split the StructureMapLayer into 2 separate classes, one for Building and the other for Construction. I will take care of it when I add Activity Spots to the Buildings display

mokun:
> Don't know about the overlapping labels

You can see two persons labels almost 'merged together' when the two are next to each other.

When that's the case, it's very hard to read.

bevans2000:
To be honest I see this as an annoyance more than than bug. Tracking the labels and having am avoidance algorithm.would be a lot of work for not enough benefit.

mokun:
> Tracking the labels... would be a lot of work for not enough benefit.

Just added a white background box with raised edge to increase the contrast and single out the selected person/robot's status info.

PR #1181, opened 2023-12-23
Rework how Persons are created

### LINKED ISSUE #47: modifying height and weight of a person by country of origin
Shall we work out a list of height modifier and weight modifier based on a person's country of origin when computing one's height and weight ?

Currently, we use the following two methods inside Person class for doing the job

> setupHeight();
> setupWeight();

In r3932, we've handcrafted a gaussian curve for setting the height/weight of a person, one for both male and female. 

However, we have yet considered one's country of origin as an important factor.
 
Also, we should look for a set of statistically accurate gender-correlated curves for setting height and weight for male and female.

e.g. from https://www.jotashoes.com/average-height-by-country, one may use the table to co-relate the height and weight of a person by one's country of origin.

Since we've already added the country field in Person, we may use it to pick a value from an array of height modifier and a value from an array of weight modifier when computing the height and weight. 

Anyone wanna help out getting this done ?

bevans2000:
This PR reworks how Persons are created to allow nation specific characteristics to be applied. The build of a Person is now self-contained within the ```Person``` constructor. The previous ```initialise``` method has been removed. 
The ```PersonBuilder``` has been updated accordingly.
There is the creation of addition Unit Tests.

This allows the Physical Characteristics of the nation to be used to create the height & weight.

Close #47


mokun:
There's a class called `Nation`. Can it be integrated with `NationSpec` ?

Instead of using `Authority`, can we use `Nation`or `NationSpec` to house all the first and last names ? 

The thing is that I was doing macro-simulation on lunar colonies. 

Currently, researchers and engineers don't have names clearly spelled out yet. Their names start with R01, R02, or E01, E02, etc. I would love to have these researchers and engineers travel between Earth, Moon and Mars.

mars-sim-core/src/main/java/com/mars_sim/core/person/NationSpecConfig.java
```diff
@@ -20,28 +21,25 @@
  */
 public class NationSpecConfig extends UserConfigurableConfig<NationSpec> {
 
-	private final String COUNTRY_XSD = "country.xsd";
+	private static final String COUNTRY_XSD = "country.xsd";
 	
 	private static final String COUNTRY = "country";
 
-	private final String ECONOMICS_DATA = "economics-data";
-//	private final String GDP = "GDP";
```
mokun, line 28: These figures (GPD, PPP, etc.) will be put into use in simulating the economic potential of a nation. 

All country xml (such as country_austria.xml) files have these 4 basic figures already.

mokun:
> https://www.worlddata.info/average-bodyheight.php

Good that you found this page.

bevans2000:
> There's a class called `Nation`. Can it be integrated with `NationSpec` ?
> 
> Instead of using `Authority`, can we use `Nation`or `NationSpec` to house all the first and last names ?
> 
> The thing is that I was doing macro-simulation on lunar colonies.
> 
> Currently, researchers and engineers don't have names clearly spelled out yet. Their names start with R01, R02, or E01, E02, etc. I would love to have these researchers and engineers travel between Earth, Moon and Mars.

No we keep names on NationSpec as the Authority can have different and the same countries. Also the bases are specific to Counties. Secondly the NationSpec only stays in memory when creating a settlement so the names do not consume long term memory storage.

I don't understand what Nation is but I will have a look. But any consolidation should be a separate issues once this is completed.

mars-sim-core/src/main/java/com/mars_sim/core/person/NationSpecConfig.java
bevans2000: They are still loaded. There is a ug because the crews are using USA but that no longer exists. I've fixed it

bevans2000:
> > https://www.worlddata.info/average-bodyheight.php
> 
> Good that you found this page.

It's a good site. I will do the other countries.

PR #1184, opened 2023-12-27
Time anomaly

mokun:
Add CPU Pulse spinner in TimeWindow for fine-grain control of pulse widith that impact ticks per seconds


bevans2000:
@mokun Are you sure this is now working correctly? I've just picked up all the changes and it seems the actual TR is not meeting the desired TR. I was using a 3 Settlement scenario.

mokun:
> the actual TR is not meeting the desired TR. 

Yay. It's catching the desired TR, sometimes larger than it and sometimes smaller than it.

I'll tweak some more.

PR #1185, opened 2023-12-28
Show multiple Settlements in the Monitor Window

### LINKED ISSUE #903: Put settlements/stations/outposts of same space agency under one belt

bevans2000:
This PR consists of changes to allow Reporting Authorities to be used in the Monitor window filter selection. The activity consists of a number of high level changes:
- Adding Authorities to the MonitorWindow combo with an indication of the number of owned Settlements. 
- When an Authority is selected; the Settlement names are shown in the toolbar of the window.
- MonitorModel converted so the selection filter is a Set of Settlements instead of a singe one
- Adding a Settlement column to most tables that is automatically shown/hidden depending on the number of Settlement being filtered.
- Category model, i.e. Food & Goods, now shown one entry per Category per Settlement

close #903

### DIFF
--- a/mars-sim-core/src/main/java/com/mars_sim/core/authority/Authority.java
+++ b/mars-sim-core/src/main/java/com/mars_sim/core/authority/Authority.java
@@ -6,12 +6,12 @@
  */
 package com.mars_sim.core.authority;
 
-import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
+import com.mars_sim.core.Entity;
 import com.mars_sim.core.configuration.UserConfigurable;
 import com.mars_sim.core.person.NationSpecConfig;
 import com.mars_sim.core.person.ai.task.util.Worker;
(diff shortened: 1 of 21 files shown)


mokun:
Sounds good !

I'll release v3.7.0 by the end of this hour. 

Will check it out.

mokun:
Now that we have a new branch 3.7.x, feel free to finish the pull !

mokun:
It works well !

bevans2000:
Yes I like it. You need a scenario with multiple Settlement on one Authority to see it best

mokun:
Ready to create weather stations, communicator stations, etc. ?

mokun:
>  I don't want to create another top-level Unit but would like to rework what we already have

Say we create a  new class called *Station*. It only needs a subset of what a full blown settlement would need to function.

bevans2000:
Do we have an issue to cover this change already?

Issue #1197, opened 2024-01-01
Unit class hierarchy inconsistency


bevans2000:
**Describe the bug**
A part of #647 the ```Unit``` hierarchy has been shown to not be accurate and the Unit class is populated with methods that are only applicable to some of the subtypes. Hence it creates an inconsistent design/model.
Unit contains methods that are only relevant to entities that can move around; e.g., setBuilding, setSettlement and more complex logic that needed to identify location state.
The reverse is also true in that classes in the ```Structure``` sub-hierarchy have to implement methods that will never be called.
This means that the Unit class has more logic than needed.

**Expected and unexpected behavior**
The solution is to create a new class ```MovableUnit``` that is a sibling on the existing ```Structure``` class that represents Unit that can move. The latter represents Unit that cannot be moved. This give a clean logic structure.
It will allow some properties/methods to be move from the Unit to the new class as well as some default implementation removed in the Structure hierarchy. 
In addition, it can strengthen the use of the ```transfer``` method as this would become abstract on the ```MoveableUnit``` class

mokun:
Yes it's good idea to create Moveable vs. Nonmoveable Unit.

bevans2000:
I think the new Unit hierarchy should be as below. Key changes:

- ```Settlement``` parented off ``Unit```
- ```MovableUnit``` new class for those Units that can move around a ```Settlement```

[...]

PR #1198, opened 2024-01-01
Delete Loggable interface

### LINKED ISSUE #647: Tweak Simlogger
- Currently, Simlogger has a long param called `long timeBetweenLogs` that will check if the same log statement is asked to be reprinted with a prescribed period of time. This param is in millisecond only and is not in msol. 
- Second, the way the log statement is printed - namely, printing it at the end and with [x84], etc., does not aid debugging.

- First, instead of using the machine time in terms of millisecond, we could provide another option of having an extra param with msol, or a convenient way of translating millisols back to miliseconds in certain cases.
- Second, printing a log statement at the beginning to report the onset of an event and then print another one at the end with [x84], etc..

 - pre-3.4.0 build 7058

bevans2000:
This changes the SimLogger to use Entity instead of Loggable for logging. Hence combining 2 interfaces into 1.

Due to the issue of #1197 the implementation is not as clean as possible. BUT Mission & ScientificStudy can now be used directly in the SimLogger statements.

Changes:
- Delete Loggable and LocationFormat classes
- SimLogger takes Entity as the class for any actor parameter
- SimLogger baselog uses the context of the Actor instead of walking the unit hierarchy
- ScientificStudy implements Entity so it can be used directly in logging statements
- Mission implements Entity so it can be used directly in logging statements
- Entity has a getContext method to
- Crop takes the context fromt the parent Building
- Mission & ScientificStudy takes the context from the lead Person's Settlement
- Structure class creates the context from the Settlement and also cached the context for direct children
- Unit getContext implements logic to dynamically identify the context based on the current container
- Added new EntityUnitTest to check context scenarios

Close #647


bevans2000:
Something very odd about these changes as it made the SonarQube Git Action freeze when running the Unit Test ```TestSaving```. I think it was something to do with the recent Pulse changes BUT oddly the normal UnitTest run works fine. So it must be a combination SonarQube and the MasterClock changes.
I have had to change how the Unit test executes the save

Issue #1204, opened 2024-01-03
Execution time anomaly


mokun:
**Describe the bug**
- On my 7th gen intel i7 machine with 8 cores, the 15-settlement default scenario template has an average execution time of 10-20 ms per frame.
- Note that the average execution time would vary from machine to machine.
- However, recent changes has caused anomalous spike of execution time as high as 5000 ms, thus more than 300x the normal amount of time it takes to complete one frame

**Stated Goal(s)**
- Benchmark the execution time within each category of Settlement's, Person's, and/or Building's timePassing() 
- Identify any unusual time loop that exceed the normal average.

**Discuss any potential problems you have and have not ruled out**
- Currently, other threads do not use that much CPU time, as shown in VisualVM.
 
**Stacktrace**
```
2024-01-03T08:35:35.0 (Config) MarsProject : Starting Mars Simulation Project 3.7.0 - Build 09e05a7-dirty - 64-bit Java 21.0.1 - 8 Cores
2024-01-03T08:35:35.0 (Config) MarsProject : List of input args : [-new, -sandbox, -timeratio, 576]
01-Adir-01:000.928 (Config) MasterClock : sleep=-380 ms, desiredTR=576, actualTR=526.54, millisol/sec=6.54, pulse/sec=14.11, mpp=70.88, ms/pulse=70.88, execution=450 ms
01-Adir-01:001.394 (Config) MarsProject : Starting the Main Window...
01-Adir-01:001.394 (Config) MainWindow : Running mars-sim in Sandbox Mode.
01-Adir-01:001.394 (Config) MasterClock : sleep=-4 ms, desiredTR=576, actualTR=483.07, millisol/sec=6.59, pulse/sec=14.19, mpp=70.48, ms/pulse=70.48, execution=74 ms
01-Adir-01:001.862 (Info) AlgaeFarming : [Tian Cheng] Algae Pond 1 - Algae fully tended for 0.04 millisols.
01-Adir-01:001.862 (Config) MasterClock : sleep=28 ms, desiredTR=576, actualTR=490.09, millisol/sec=6.59, pulse/sec=14.16, mpp=70.60, ms/pulse=70.60, execution=42 ms
01-Adir-01:002.330 (Config) MasterClock : sleep=47 ms, desiredTR=576, actualTR=498.07, millisol/sec=6.58, pulse/sec=14.13, mpp=70.75, ms/pulse=70.75, execution=23 ms
[...]

Issue #1192, opened 2023-12-31
Excessive CPU on doAreasCollide


mokun:
**Describe the bug**
- During the day, settlers go outside. The most CPU intensive ( 47.8 + 21% ~70%) or punishing task turns out to be checking for collision when the task `WalkOutside` is enacted.  

**Expected and unexpected behavior**
- It's too expensive to check if two areas are collided when a person is walking outside. 
- It's even possible for a person to climb up a building. So collision checking is not needed.
- It only needs to look at the x/y center location of a person with other objects in the Settlement map.
- A person is a dot and occupies negligible area on the settlement map. It's not like a rover that has a much large footprint.

**Java classes involved**
- LocalAreaUtil, WalkOutside

**Screenshots**
![image](https://github.com/mars-sim/mars-sim/assets/1168584/d9d2f016-6b06-47c3-b97a-16cd7b02dae6)

**Specifications**
- post 3.7.0 build 9172

**Additional context**
- Add any other context about the problem here.

mokun:
@bevans2000 

What do you make of this ? 

Can you by any chance fire up visualvm to see if there's anything out of ordinary that eat up CPU ?

bevans2000:
Ok, probably not u til the weekend as busy with work.

mokun:
@bevans2000 

I wonder if defining an exact spot for each settlement to dig local regolith / ice is better.

This way, we may be able to cache that spot and its results when checking for collisions by  repeatedly using the A* path planning algorithm on cached locations.

bevans2000:
I don't think this will help this problem but from a logic prese give it make sense. 
Write some Unit tests to test the path calculations

mokun:
`WalkOutside`, plus all other walking tasks, are called to be a subtask only.

Does a time pulse get sent to a subtask the same way as to a main task ? 

Also, it doesn't seem to see the progression of time on anything one performing a walking task in the last couple version of mars-sim anymore.

It's possible subtasks used to be handled more carefully in terms of time usage but not now.

So the question is, how do we modify `WalkOutside` to ensure that a subtask like that is being regulated by our time pulse architecture again ?

bevans2000:
Well when I redesigned the rask execution a long time ago subtasks were treated the same and would have seen the increasing time via a pulse. But there have been other changes like the standard pulse width that you overlaid on top.

mokun:
This seems to be how the A* pathfinding search algorithm is implemented. 

[...]

Issue #655, opened 2022-08-03
Rework MasterClock's time ratio to improve stability


mokun:
**Is your feature request related to a problem? Please describe.**
- Rework MasterClock's *time ratio* to increase NOT as the exponent of 2.

**Describe the solution you'd like**
- May consider starting at 128x in the multiple of 128x for increase, such as 128x. 128x, 256x, 512x, 768x, 896x, 1024x, 1152x, 1280x, 1408x, 1536x, etc.
- Or starting at 256x in the multiple of 256x.

**Describe alternatives you've considered**
- Make various functions such as `Heating` work better under high time ratio in long run.
- Only change the time ratio increase after 1024x.

**Specifications  (please complete the following information):**
- pre-3.4.0 build 7089
 
**Additional context**
- Increase the ***pulse width*** steadily as time ratio increases in order to prevent negative sleep time.

mokun:
@bevans2000 

I'm experimenting on allowing a gradual increase of the pulse width, aka `lastPulseTime` in MasterClock, in relation to the time ratio.

Do you have any concerns over that ?


How we currently model `lastPulseTime ` is as follows: 

```
				// Get the time pulse length in millisols.
				lastPulseTime = (realElaspedMilliSec * desiredTR) / MILLISECONDS_PER_MILLISOL;

				// Pulse must be less than the max and positive
				if (lastPulseTime > 0) {

					acceptablePulse = true;

					if (lastPulseTime > maxMilliSolPerPulse) {
						logger.config(20_000, "Pulse width " + Math.round(lastPulseTime*100_000.0)/100_000.0
								+ " clipped to a max of " + maxMilliSolPerPulse + ".");
						lastPulseTime = maxMilliSolPerPulse;
					
					}
					else if (lastPulseTime < minMilliSolPerPulse) {
						logger.config(20_000, "Pulse width " + Math.round(lastPulseTime*100_000.0)/100_000.0
								+ " increased to a minimum of " + minMilliSolPerPulse + ".");
						lastPulseTime = minMilliSolPerPulse;
					}
				}
```

mokun:
I would expect the current *pulse width* should be compared to the `optMilliSolPerPulse` at the end of each frame. 

So if the *pulse width* deviates too much from it, then one would gradually bring it back to `optMilliSolPerPulse`.

```
		minMilliSolPerPulse = simulationConfig.getMinSimulatedPulse();
[...]

PR #1207, opened 2024-01-05
Adjust A* pathfinding algorithm

mokun:
1.4.2024

Note: this commit seems to eliminate execution time spike due to A* pathfinding algorithm implementation

- new: add count param to track the while loop iteration in determineObstacleAvoidancePath() in WalkOutside
- change: limit count to eliminate execution time spike

Relates to #655, #1192 and #1204

### DIFF
--- a/mars-sim-core/src/main/java/com/mars_sim/core/person/ai/task/DigLocal.java
+++ b/mars-sim-core/src/main/java/com/mars_sim/core/person/ai/task/DigLocal.java
@@ -487,7 +487,10 @@ public abstract class DigLocal extends EVAOperation {
 		}
 		int rand = RandomUtil.getRandomInt(binList.size() - 1);
         Building b = binList.get(rand); 
-		
+		if (b != null) {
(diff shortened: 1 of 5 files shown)


mokun:
@bevans2000 

I'm going to need to comment out check if outsideWalk.isDone() in line 146 in WalkOutsideTest.

This test has been problematic here and there in the GitHub test. In my machine maven test, it throws up once in a while. 

Now that we limit the while loop iteration to prevent execution time spike, it may affect the passing of this test once in a while.

But in simulation session, all settlers can walk to a digging location and to a bin location and walk back. I don't see any problems.

bevans2000:
> @bevans2000
> 
> I'm going to need to comment out check if outsideWalk.isDone() in line 146 in WalkOutsideTest.
> 
Short term ok but we need to increase the unit test coverage so we can find these logic problem outside the simulation. The path finding of the WalkingSteps class should be fully Unit tested. The actual walking Task in a Task and they are difficult to test. I want to create a harness so we can test Task execution in a Unit test. I think we can simulate the pulse going to the Mind class.

PR #1214, opened 2024-01-07
Stop the selling & buying list having the same Good

### LINKED ISSUE #1178: Goods to buy and sell
- Currently, it's not difficult to find the same good in both *Goods to Sell* Table and *Desired Goods to Buy* table in a Trading or Delivery mission.

**Describe the rationale**
- Avoid buying and settling the same goods

- Find out if the criteria for buying and settling certain goods are working
- Compare its string name, instead of its Good object reference), since it's across different threads and the object reference of the same good on two settlements are not the same

**Screenshots**
image

image

image

**Specifications**
- pre 3.7.0 build 9107

**Additional context**
- See https://github.com/mars-sim/mars-sim/issues/506

bevans2000:
Deals have the same Goods in the to sell and to Buy list which make no sense. This change prevents, for each Settlement, a Good in the buying list will not be added to the selling list. The buying list takes priority.

Changes:
- GoodManager adds the Good in the buying list to the excluded Goods when creating the selling list.
- Clean code smells in Good class

Close #1178

### DIFF
--- a/mars-sim-core/src/main/java/com/mars_sim/core/goods/Good.java
+++ b/mars-sim-core/src/main/java/com/mars_sim/core/goods/Good.java
@@ -8,13 +8,13 @@ package com.mars_sim.core.goods;
 
 import java.io.Serializable;
 import java.util.HashMap;
-import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.stream.Stream;
 
 import com.mars_sim.core.SimulationConfig;
+import com.mars_sim.core.Unit;
 import com.mars_sim.core.food.FoodProductionProcessInfo;
 import com.mars_sim.core.food.FoodProductionProcessItem;
 import com.mars_sim.core.food.FoodProductionUtil;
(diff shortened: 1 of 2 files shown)


mokun:
@bevans2000, 

Can we merge these changes onto 3.7.x as well since it's considered a bug fix ? 

It's not like it's adding any new features.

bevans2000:
Well I don't think the impact of the problem justifies the bugfix. Its bit that easy because it may pick up a lot of other changes. You round have to cherry pick the commit

mokun:
> Well I don't think the impact of the problem justifies the bugfix. 

I don't understand that.

> Its bit that easy because it may pick up a lot of other changes. You round have to cherry pick the commit

have to cherry pick the commit ? what does it mean ?

You only changed 2 files right ?

I'll simply copy and paste the new contents to the 3.7.x branch.

bevans2000:
The issue was marked for 3.8.0 so that's is where it is fixed.
You could double fix it and copy the 1 file (it's only GoidsManager) but you will then gave a merge conflict when you merge your bugfux branch back to master.
Search for git cherry picking commits.

PR #1216, opened 2024-01-09
Fix location update

mokun:
1.8.2024

- new: add vicinityUnit and vicinityLabel in LocationTabPanel
- change: revise findSettlementVicinity() in LocationTag
- change: remove settlement container in LocationTabPanel
- fix: correct findSettlement() in UnitManager
- fix: correct a few methods in LocationTag when finding the settlement or vehicle vicinity
- fix: correct finding the location of a person, robot, equipment, vehicle in the vicinity of a settlement or vehicle


bevans2000:
@mokun  There is an error in the UnitTest now. A null pointer is thrown but is trapped somewhere so the test is not reporting an error

Screenshot 2024-01-09 220702

mokun:
It may have to do with this line in MasterClock on Lin 867 : 

> Collections.synchronizedSet(new HashSet<>(clockListenerTasks)).stream().forEach(this::executeClockListenerTask);

Don't know why it throws NPE ? 

Why is "c" null ?

if *clockListenerTasks* is empty, the HashSet should be empty and shouldn't be null.

PR #1254, opened 2024-02-11
Merge Maven main module into ui module

bevans2000:
This PR consists of 3 major changes:

1. Update version tag to '3.8.0-pre'
2. Clean the dependencies and plugin for the Maven module to remove unused references
3. The UI module absorbs main module such that it creates the executable JAR for the Swing version

The 'main' module folder is till left but this can be removed.

Note, this change may need a reload of the project in Eclipse. VSCode is fine.


pom.xml
```diff
@@ -148,7 +147,6 @@
 	<modules>
     	<module>mars-sim-tools</module>
 		<module>mars-sim-mapdata</module>
-		<module>mars-sim-main</module>
```
mokun, line 155: So, the maven submodule mars-sim-main have no use now ?

mokun:
That's a big change !


I wonder if you still want `mars-sim-tools`.

pom.xml
bevans2000: Correct

bevans2000:
> That's a big change !
> 
> I wonder if you still want `mars-sim-tools`.

Yes, this is the first change. Next is to combine tools, map data and core. 
Then tools will just contain our own tools like the HTML generator
I want to make sure this change works for you with Eclipse

mokun:
> I want to make sure this change works for you with Eclipse

Yes. After I choose "Update Project" under "Maven", things got sync-ed. It works.

mokun:
> VSCode is fine

Somehow I find it uncomfortable to develop mars-sim in IntelliJ.

But I'd love to see how easy mars-sim can be set up for development in VSCode. 

This way, I don't have to depend on Eclipse forever.

bevans2000:
> > VSCode is fine
> 
> Somehow I find it uncomfortable to develop mars-sim in IntelliJ.
> 
> But I'd love to see how easy mars-sim can be set up for development in VSCode.
> 
> This way, I don't have to depend on Eclipse forever.

VSCode is pretty good. I switched over a year ago. I found Eclipse was getting slower.

Will consolidate the core module next. Overall,this will simplify the build.

PR #1263, opened 2024-02-17
Remove mortar in ResourceUtil

mokun:
02.17.2024


mokun:
I'll need to update all the eclipse launcher to point to the MarsProject class correctly.

mokun:
What do you plan to do with `mars-sim-headless` ? 

Will it be deleted as well ?

PR #1284, opened 2024-04-06
Create a Group Activity tab to the Settlement window

### LINKED ISSUE #1282: Standardise Unit Tab Panel windows
The unit windows consist of a number of tab panels showing different aspects on a Unit. The tabs use a mixture of styles to display the information. The majority present a list of entities.
The tab panels have a lot of duplicated code.

Provide a standard approach for displaying a list of entities and an associated fixed information panel as a single tab.
This can also replace the use of UnitList with a set of standard models, e.g Worker showing current Task or 
One goal is to significantly reduce repeated code and logic. As well as having a standard approach.

Consolidate the layout and construction of the table and behaviour into a new class. The class would be a Tab panel but provide logic to built a table and control it's behaviour. This class would be subclassed for specific aspects on a Unit where the responsibility would be to provide a Table model. 
Optionally subclasses can also provide an information panel to render above the table.

bevans2000:
This PR adds a new tab panel to the Settlement Unit window to show any Group Activitites.

To deliver this there has been a standardisatino of the tab panels. Many tab panel have the same pattern of an info panel with a table below.
This UI design is now captured as a new class ```TabPanelTable``` that centralises teh layout and construction of the panel.
The specific TabPanel just needs to problem a table model as a minumum. If that model implement the ```UnitModel``` interface then a Unit launcher will be automaticalyl added.

This pattern can be further extended to remove the use of the flat list ```UnitListPanel``` and convert those panels to the same approach.
The addition of a standard Person model with selectable column would be beneficial; based on the existing PersonTableModel used in the MonitorTool.

Close #1282

### DIFF
--- a/mars-sim-core/src/main/java/com/mars_sim/core/activities/GroupActivityMetaTask.java
+++ b/mars-sim-core/src/main/java/com/mars_sim/core/activities/GroupActivityMetaTask.java
@@ -60,11 +60,7 @@ public class GroupActivityMetaTask extends MetaTask implements SettlementMetaTas
      */
     @Override
     public List<SettlementTask> getSettlementTasks(Settlement settlement) {
-       var active = settlement.getFutureManager().getEvents().stream()
-                .filter(e -> e.getHandler() instanceof GroupActivity)
-                .map(e -> (GroupActivity)e.getHandler())
-                .filter(GroupActivity::isActive)
-                .toList();
+        var active = settlement.getGroupActivities(true);
         
          List<SettlementTask> results = new ArrayList<>();
          for(var a : active) {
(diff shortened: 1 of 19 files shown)


mokun:
It's good to show the available group activities for each settlement.

We should also definitely provide UI controls for players to create new ones as well. Shall we ?

mokun:
One thing though, I do feel the settlement unit window tab already a bit too overwhelmimg. Do you ?

When it's time we are going to design the entire main window one day, we need to go beyond the logical topology of giving each Unit an Unit Window.

We could create various new UI tools to highlight certain authorities entitled to a commander to take hold of. 

He can directly intervene in any situations or in normally day-to-day  work, he can create new plans and initiatives...

This will reduce the frontal overload of giving players all kinds of statistics.

Any new UI tools should be about what a commander CAN DO for a settlement.

His decision should have huge impact, not like right now the commander doesn't have that much influence on the course of the settlement 's future.

bevans2000:
> One thing though, I do feel the settlement unit window tab already a bit too overwhelmimg. Do you ?
> 
Yes. Data that is already available via the Monitor tool should not be repeated as a tab, e.g vehicles, citizens

PR #1314, opened 2024-06-09
Update Muscle stress via Experience Impact

### LINKED ISSUE #1285: Simplify and consolidate how a Worker is impacted by doing a Task/Activity
The Task class and it's subclasses have numerous places where the Task impacts the properties of a Worker, e.g. skill improvement, fatgiue, stress.
This is complex and hard to understand and has very low encapsulation. Also sometimes there is double counting occurring and no consistency.
Also the new ```GroupActivity``` class needs a means of defining the benefits of a Person attending an activity (see #971)

Capture the impact of doing "something" for a period of time as a single class. This would update all the relevant properties of a Worker as one action based on a set of properties.
This class will be stateless allowing it to be shared by multiple Tasks/activities or in future Missions as well.
It will also provide a singular means to defining impact.
This could be loaded from a configuration file as well.

Create a new stateless class called ```ExperienceImpact```. This will have a method ```apply``` that Workers call for a time period to update their properties. This represents them experiencing this activity for a certain time duration.
The experience class will be created a a static for most Task/GroupActivity instances allowing it to be centralised and shared by all instances of that Task type.
Formula to assess the impact can be centralised based on a single set of parameters. There needs to be the ability to enhance the calculations based on any external factors the Work is experiencing such as teaching.
The Task fatigue, stress, skill & emotion modifications can be centralised into a single logic flow.

### LINKED ISSUE #1311: Manage Person Muscle stress in Task
**Describe the bug**
Some Tasks apply a muscle stress to the Person doing the work when there is heavy physical effort needed. For example Unloading a vehicle. These are called separately and individually instead of centralised meaning the fully impact is not clear.
 
Enhanced the ```ExperienceImpact``` class to apply muscle stress to a Person as part of the existing 'apply' pattern. This method is already responsible for all the impact a Task has on a Worker.
To implement this a new enum will be created that defines the effort needed NONE, LOW, HIGH. The enum will be passed in the constructor replacing the existing Boolean representation.

bevans2000:
This PR moves how muscle stress is applied during Tasks. 

This is delivered with the following changes:
- ExperienceImpact categorises the Physical Effort as None, Low, High
- Applying an impact with a High effort will stress the muscle of the Person worker

The Loading & Unloading Tasks are classified as High effort.
In addition, the Read Task is modified to adopt the pattern of #1285

close #1311
close #1285


bevans2000:
There is a little instability in a small number of new Task Unit Tests I have notice. It's due to the randomness. I will eventually track them down and make them stable.

PR #1343, opened 2024-07-07
Fix Building Category

### LINKED ISSUE #1339: Wrong building category
**Describe the bug**
- Currently, building such as Garage, Astronomy observatory, hallway, etc. are assigned with wrong building category

- BuildingConfig should intelligently find the correct category for each building, instead of having to manually assign one in buildings.xml. 

**Java classes involved**
- BuildingConfig, buildings.xml, BuildingCategory 
- In BuildingConfig,
```java
	/**
	 * Derives the category from the types of Functions.
	 * 
	 * @param functions
	 * @return
	 */
	private static BuildingCategory deriveCategory(Set<FunctionType> functions) {

		// Get a set of categories
		Set<BuildingCategory> cats = functions.stream()
						.map(f -> f.getCategory())
						.collect(Collectors.toSet());

		BuildingCategory category = BuildingCategory.CONNECTION;
		if (!cats.isEmpty()) {
			// Find the category with the lowest ordinal as that is the best to represent
			// this set of Functions
			int lowestOrdinal = 999;
			for (BuildingCategory c : cats) {
				if (c.ordinal() < lowestOrdinal) {
					lowestOrdinal = c.ordinal();
					category = c;
				}
			}
		}
		return category;
	}
```

**Screenshots**
- Garage should belong to workshop or vehicle maint category 
image

- Note that the Living Space category has already been fixed in recent commit
image

**Specifications**
- pre-3.8.0

**Additional context**
- The use of ordinal in BuildingCategory doesn't seem to bode well.
- Workshop is NOT a laboratory. It should have a new category called workshop or something else
- In BuildingCategory,
```java
public enum BuildingCategory {

    // Note the order of these enums represent their importance
    COMMAND, 
    COMMUNICATION,
    MEDICAL,
    ERV,
    FARMING, 
    ASTRONOMY,
    LABORATORY, 
    LIVING,
    PROCESSING,
    WORKSHOP,
    VEHICLE,
    STORAGE, 
    EVA,
    CONNECTION,
    POWER;
```
- Note the attribute `category="CONNECTION"` was added recently in buildings.xml to help defining what category a hallway should belong. Before that, it was under LIVING. Now, it's CONNECTION.
```xml
	<!-- A hallway's length is variable and determined in templates or during construction. -->
	<building type="Hallway" width="2.0" length="-1.0" north-south-alignment="length" 
			construction="SEMI_SOLID" base-level="0"
			wear-lifetime="3340000" maintenance-time="20" room-temperature="20.0" base-mass="200"
			category="CONNECTION">
```

bevans2000:
Issue #1339 highlighted some Buildings are not getting the correct category. The problem lies in the LifeSupportt & RoboticStation Functions have a Living category assigned. These are always present and hence override any other Functions. The correction is to allow a FunctionType to have no category, like the 2 above. Then these are excluded from the Category logic.

Changes:
- Remove category from Life Supprot & RoboticStation
- BuildingConfig can handle a Functiontype not having a category
- Create a new UnitTest to check category of some common buildings

Fix #1339


mars-sim-core/src/test/java/com/mars_sim/core/structure/building/BuildingCategoryTest.java
```diff
@@ -0,0 +1,47 @@
...
+    }
+   
+    public void testCommand() {
+        assertBuildingType("Command and Control", BuildingCategory.COMMAND);
+    }
+    
+    public void testQuarters() {
+        assertBuildingType("Residential Quarters", BuildingCategory.LIVING);
+    }
+
```
mokun, line 38: Let's also add testWorkshop(). 

We'll need to fix the order of WORKSHOP to go before LABORATORY in BuildingCategory.

mokun (approved):

PR #1350, opened 2024-07-13
Vehicle Crew tab panel manages members better

bevans2000:
The Vehicle Crew tab handles the Unit management in a better logic. Also the UnitWindow destroys the tabs when the window is closed. This ensures that the tabs are deregistering any listeners when the window is closed.

In addition, this PR contains a change to add a Remove button to the Preferences tab


mokun:
> add a Remove button to the Preferences tab

That's awesome !

PR #1398, opened 2024-09-01
Create a SuppliesManifest

bevans2000:
This PR reworks the ```MissionManifest``` to a more generic ```SuppliesManifest```. The latter allows Amounts, Equipment & Items to be separately added to a manifest as either mandatory or optional.

The ```LoadingController``` uses the manifest as the source of the loading instead of a combination of separate Maps. The controller also explicitly manages the loading of Items & Amounts.

The objective is to improve the clarity of the code and simplify the loading.


bevans2000:
@mokun The CodeFactor check is failing but it does not pick up  all issues resolved for some reason.

PR #1404, opened 2024-09-07
Extract new method from duplicate, introduce boolean for more readability

muybien3433:
PR refactors duplicate code by extracting a new method and introduces a boolean to improve readability.

### DIFF
--- a/mars-sim-core/src/main/java/com/mars_sim/core/air/AirComposition.java
+++ b/mars-sim-core/src/main/java/com/mars_sim/core/air/AirComposition.java
@@ -167,8 +167,9 @@ public class AirComposition implements Serializable {
 	public void timePassing(Building building, ClockPulse pulse) {
 		double time = pulse.getElapsed();
 		double tt = building.getCurrentTemperature();
+		boolean isTempUnsafe = tt > -40 && tt < 40;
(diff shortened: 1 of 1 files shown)


mars-sim-core/src/main/java/com/mars_sim/core/air/AirComposition.java
```diff
@@ -411,43 +412,54 @@ public static void initializeInstances(PersonConfig personConfig) {
...
+
+	/**
+	 *
+	 * Gets mass or pressure based for particular gas.
+	 *
+	 * @param gasId
+	 * @return
+	 */
+	private static double getByGas(int gasId, double co2,
+								   double argon, double n2, double o2, double h2o) {
```
bevans2000, line 451: @muybien3433  Nice approach. This code has been bugging me for a while.

bevans2000 (approved):

bevans2000:
Interesting that there must be a problem in our GitHub action configuration. As the source branch is outside the repo, it can not access the secret for the SonarCloud credentials. Must be a way to resolve that; seems the action execute in the context of the remote repo not the target repo.
But not the end of the world at the moment.

PR #1357, opened 2024-07-21
Old: Remove dependency on Mission for Loading/Unloading a Vehicle

bevans2000:
This PR removes the dependency on Mission for Loading/Unloading a Vehicle. These operations are now triggered by using new dedicated StatusTypes values. 
In addition the Loading is now instigated by using a ```SuppliesManfiest``` instead of creating a LoadingController. The latter is now a private class to the Vehicle to manage the loading. Requestors of a load have to only specific the supplies to be loaded by a SuppliesManifest.

To support this the definition of Parked vehicles at a settlement has been changed; it now incudes Vehicles visiting from other Settlements.

Issue #1324, opened 2024-06-16
Change triggering of Vehicle Loading/Unloading


bevans2000:
**Is your feature request related to a problem? Please describe.**
Vehicle loading & Unloading is directly triggered by a Vehicle Mission. This make the logic complicated and not easy to test in isolation.
In addition the Delivery & Trade Missions have further complication is have to perform a load/unload at the remote Settlement.

**Describe the goal(s) you have in mind**
The loading/unloading of a Vehicle is an activity specific to a vehicle and not a Mission.  It is possible in the future there maybe cases where a Vehicle needs loading outside a mission, for example training.

**Describe the solution you desire to see**
The change is to use a new Secondary Vehicle status to trigger the loading or unloading of a Vehicle. The Meta Task would look for any Vehicle parked at a Settlement that has the appropriate status instead of searching all Mission currently.
The Mission would indirectly trigger the load/unload by setting the appropriate status and wait for the status to be cleared. This significantly simplifies the Trade & Delivery Missions.

mokun:
> possible in the future there maybe cases where a Vehicle needs loading outside a mission, for example training

I like that observation. 

I can think of driving a vehicle to a regolith bin to load and unload regolith. Currently it's difficult to do unless we work out a custom solution.

bevans2000:
Yes good example.

mokun:
Is this done ?

PR #1405, opened 2024-09-07
Loading of vehicles is independent of VehicleMission

bevans2000:
This PR makes changes so the loading of a Vehicle is independent of a VehicleMission. This reduces the coupling of code and allow future change to load Vehicles without needing a mission. 

This is delivered by introducing a new StatusType of LOADING this signifies Vehicles ready for loading. Secondary benefit is the LOADING is recorded in the status log of the Vehicle.
When in the LOADING state, a Vehicle now has a ```LoadingController``` attached instead of it being on the ```VehicleMission```.

To support this, the definition of a parked vehicle is changed to include those visiting from other Settlements.

Complementary changes for unloading will follow.
This supersedes the PR #1357 which proved impossible to merge changes due to significant changes on the master branch.

In addition, ```DustStorm``` now supports a description instead of relying upon the Settlement.

Part of #1324

### DIFF
--- a/mars-sim-console/src/main/java/com/mars_sim/console/chat/simcommand/CommandHelper.java
+++ b/mars-sim-console/src/main/java/com/mars_sim/console/chat/simcommand/CommandHelper.java
@@ -373,15 +373,18 @@ public class CommandHelper {
 			}
 	
 			// Vehicle mission has a loading
-			LoadingController lp = tm.getLoadingPlan();
-			if ((lp != null) && !lp.isCompleted()) {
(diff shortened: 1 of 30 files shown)


mokun:
> DustStorm now supports a description instead of relying upon the Settlement.

Yes and we'll need to consider how we may pin a dust storm or draw out a storm front on a global map in future. 

Obviously, it would be interesting to see some sort of weather maps and to see  how a dust storm moves across the map  in within period of time, much like I'm a weather report.

Issue #1365, opened 2024-08-04
Unable to start new missions


mokun:
**Is your feature request related to a problem? Please describe.**
- There are situations in which missions can no longer be started after a few sols.
- It seems after loading from a saved simulation, none of the settlers in a settlement is able to start new missions.
- It is unknown as to why calling Mind's startNewMission() would result in null ALL THE TIME in PlanMission's selectingPhase().

**Describe the rationale**
- The benefit of showing the mission rating probability enable developers to debug why missions cannot be started.
- It also enables players to understand why certain mission types are more favorable than others in a settlement.

**Describe the goal(s) you have in mind**
- Disclose the mission rating probability scores that influence how a person may choose to start a new mission, just like showing the task rating probability in Activity tab in PersonUnitWindow.

**Describe the solution you desire to see**
- Add a table in Activity tab in PersonUnitWindow for viewing the mission rating scores for each mission by a settler.

**Screenshots**
- In sol 36 already. No mission generated after sol 6. 
- The last mission returned in sol 9. No more missions afterward.
![image](https://github.com/user-attachments/assets/849bdd30-f443-4236-87ac-1beeb31dbb34)

- Currently, Activity tab in PersonUnitWindow show the task ratings and how a person would select a task.
- Similarly, the mission ratings should be displayed.
![image](https://github.com/user-attachments/assets/02a1fe93-afb7-483c-ae2e-6b6012bb9a60)

**MarsLink**
```
[01-Bora-08:728 Sol 36] MarsLink AI >  ​/mp
Show me the statistics on the mission plans submitted.

 1. Today
 2. last 3 sols (Each)
 3. Last 7 sols (Each)
 4. last 3 sols (Combined)
 5. Last 7 sols (Combined)
 6. Last 14 sols (Combined)
 7. Last 28 sols (Combined)
 8. Since the beginning (Combined)

Enter your choice (1-8)
Select an option  ​7
No plans submitted.

[01-Bora-08:728 Sol 36] MarsLink AI >  ​/ms
                    Name |                Phase |                Vehicle |             Settlement
-------------------------------------------------------------------------------------------------
Regolith Prospecting 2-001 |            Completed |          Pathfinder II |          Sagan Station
Regolith Prospecting 2-001 |            Completed |                Agartha |          Sagan Station
Regolith Prospecting 2-002 |            Completed |                Atlanta |          Sagan Station
Regolith Prospecting 3-001 |            Completed |                Liberty |          Sagan Station
[...]

PR #1428, opened 2024-09-12
Create new Generic Class to use with mission and task

muybien3433:
Following issue #1365 
I created a new generic class to address the duplicated methods. Let me know what you think

### DIFF
--- a/mars-sim-core/src/main/java/com/mars_sim/core/person/ai/CacheCreator.java
+++ b/mars-sim-core/src/main/java/com/mars_sim/core/person/ai/CacheCreator.java
@@ -1,19 +1,19 @@
 package com.mars_sim.core.person.ai;
 
-import com.mars_sim.core.person.ai.task.util.TaskJob;
+import com.mars_sim.core.data.Rating;
 import com.mars_sim.core.time.MarsTime;
 import com.mars_sim.core.tool.RandomUtil;
 
 import java.util.ArrayList;
 import java.util.List;
 
-public class CacheCreator<T extends TaskJob> {
+public class CacheCreator<T extends Rating> {
 
     private List<T> cache = new ArrayList<>();
     private double probability = 0;
     private String context;
     private MarsTime createdTime;
-    private TaskJob lastSelected;
+    private T lastSelected;
 
     public CacheCreator(String context, MarsTime createdTime) {
         this.context = context;
@@ -52,7 +52,7 @@ public class CacheCreator<T extends TaskJob> {
         return cache;
     }
 
-    public TaskJob getLastSelected() {
+    public T getLastSelected() {
         return lastSelected;
     }
 
@@ -80,10 +80,13 @@ public class CacheCreator<T extends TaskJob> {
         return RandomUtil.getRandomDouble(probability);
     }
 
+
+
     public void destroy() {
         cache.clear();
         cache = null;
         createdTime = null;
         lastSelected = null;
     }
+
 }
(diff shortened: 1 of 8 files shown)


mars-sim-core/src/main/java/com/mars_sim/core/person/ai/CacheCreator.java
```diff
@@ -0,0 +1,89 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+public class CacheCreator<T extends TaskJob> {
+
+    private List<T> cache = new ArrayList<>();
+    private double probability = 0;
+    private String context;
+    private MarsTime createdTime;
+    private TaskJob lastSelected;
```
bevans2000: I think this should really be a type T?

mars-sim-core/src/main/java/com/mars_sim/core/person/ai/CacheCreator.java
```diff
@@ -0,0 +1,89 @@
...
+            lastEntry = entry;
+        }
+        return lastEntry;
+    }
+
+    private double getRandomDoubleBasedOnProbability() {
+        return RandomUtil.getRandomDouble(probability);
+    }
+
+    public void destroy() {
```
bevans2000, line 85: Hate these destroy methods, they are a bad anti-pattern

mars-sim-core/src/main/java/com/mars_sim/core/person/ai/CacheCreator.java
```diff
@@ -0,0 +1,89 @@
+package com.mars_sim.core.person.ai;
+
+import com.mars_sim.core.person.ai.task.util.TaskJob;
+import com.mars_sim.core.time.MarsTime;
+import com.mars_sim.core.tool.RandomUtil;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class CacheCreator<T extends TaskJob> {
```
bevans2000: Can we expand it to also replace the Mission selection code? We would a new interface, ScoredItem, that has a get score method.
Then we could use the Generics for both cases. Maybe rename the class to ScoredItemChooser.

mars-sim-core/src/main/java/com/mars_sim/core/person/ai/CacheCreator.java
```diff
@@ -0,0 +1,89 @@
+package com.mars_sim.core.person.ai;
```
bevans2000, line 1: Can we create a simple unit test to check the selection logic?

mars-sim-core/src/main/java/com/mars_sim/core/person/ai/CacheCreator.java
mokun: But we want that object destroy right away to minimize the disruption of the garbage collector, right ?

mars-sim-core/src/main/java/com/mars_sim/core/person/ai/CacheCreator.java
mokun: Have to give some thoughts on how it can be tested. 

@muybien3433 

Here are the [examples](https://github.com/mars-sim/mars-sim/tree/master/mars-sim-core/src/test/java/com/mars_sim/core/person/ai/task) of the tests that we'll modify after the change is made. 

Now you know where they are. Namely, inside the *'/src/test/* directory in each package.

muybien3433: I’ll take care of this after weekend.

PR #1435, opened 2024-09-16
Add mockito to maven dependency, create CacheCreatorTest

muybien3433:
I find that tests written with Mockito are more straightforward and easier to read.


bevans2000:
@muybien3433 yes let's give mockito a try. We generally have a mixture of styles in the unit test, mainly are JUnit 4. We need to refresh them all to confirm to JUnit 5 and use annotations

Issue #1356, opened 2024-07-21
Crewmembers teleported. Vehicle stuck. A Crewmember having a medical emergency 


mokun:
**Describe the bug**
- Crewmembers teleported back to their home settlement.
- Vehicle stuck and unable to move.
- It seems that a Crewmember was having a medical emergency that causes all these troubles.

**Expected and unexpected behavior**
-  It should be able to help curtailing the mission so that vehicle may return home quickly, if a crewmember has a medical emergency.

**Java classes involved**
- RoverMission, AbstractVehicleMission

**Screenshots**
- Pyotr had a medical emergency. Mission halted. 
- Vehicle should have immediately aborted the mission and start driving back.
- Vehicle is reportedly seen at settlement vicinity.
![image](https://github.com/user-attachments/assets/21472a76-a694-46c4-b2b6-cbd604abc368)

- All four of them got "teleported" back to the vicinity of the settlement
![image](https://github.com/user-attachments/assets/b3299bab-f597-4334-aebb-836719e8c028)

- Rover Middendorf never gotten off the mission list.
- Middendorf is not seen anywhere in Settlement Map yet.
![image](https://github.com/user-attachments/assets/8fd32bec-c061-4e32-a6b4-23a3ee65c14e)

**Stacktrace**
```
01-Adir-03:270.317 (Info) Walk : [Zvezda - Garage 1 - Middendorf] Pyotr Volkov - Entered rover Middendorf inside Garage 1.
01-Adir-03:270.317 (Info) EVASuitUtil : [Zvezda - Garage 1 - Middendorf] Pyotr Volkov - Just fetched EVA Suit 184 from Zvezda to Middendorf.
01-Adir-03:270.317 (Info) EVASuitUtil : [Zvezda - Garage 1 - Middendorf] Valeria Kuznetsov - Just fetched EVA Suit 169 from Zvezda to Middendorf.
01-Adir-03:270.317 (Info) EVASuitUtil : [Zvezda - Garage 1 - Middendorf] Alexander Alexeev - Just fetched EVA Suit 180 from Zvezda to Middendorf.
01-Adir-03:270.317 (Info) EVASuitUtil : [Zvezda - Garage 1 - Middendorf] Fyodor Lebedev - Just fetched EVA Suit 162 from Zvezda to Middendorf.
...
01-Adir-03:270.317 (Info) Vehicle : [Zvezda] Middendorf - Moved to new parking loc at (-7.0, -15.7).
[...]

PR #1436, opened 2024-09-17
Remove duplicates, extract new class and methods

muybien3433:
I created a new package called 'common' to manage duplicates in the vehicle package. I've rewritten one, but there are more in other classes that I'll handle next. 

Is there any reason to leave comments like this?:
  /** The VehicleSpec instance. */
	private VehicleSpec spec;
	/** The mission instance. */
	private Mission mission;

Pretty self-explanatory just like this:
	/**
	 * Gets the vehicle type.
	 * 
	 * @return
	 */
	public VehicleType getVehicleType() {
		return vehicleType;
	}

Many of them can be removed simply by changing method name to reflect what the comment says.

### DIFF
--- a/mars-sim-core/src/test/java/com/mars_sim/core/person/ai/CacheCreatorTest.java
+++ b/mars-sim-core/src/test/java/com/mars_sim/core/person/ai/CacheCreatorTest.java
@@ -29,7 +29,6 @@ class CacheCreatorTest {
     @BeforeEach
     void setUp() {
         MockitoAnnotations.openMocks(this);
-        cacheCreator = spy(new CacheCreator<>("test-context", marsTime));
 
         when(score1.getScore()).thenReturn(5.0);
         when(score2.getScore()).thenReturn(3.0);
--- a/mars-sim-core/src/test/java/com/mars_sim/core/person/ai/CacheCreatorTest.java
+++ b/mars-sim-core/src/test/java/com/mars_sim/core/person/ai/CacheCreatorTest.java
@@ -1,0 +1,13 @@
+package com.mars_sim.core.person.ai;
+
+import com.mars_sim.core.data.Rating;
+import com.mars_sim.core.time.MarsTime;
+import org.mockito.Mock;
+
+class CacheCreatorTest {
+
+    private CacheCreator<Rating> cacheCreator;
+    private MarsTime marsTime;
+
+    }
+}
--- a/mars-sim-core/src/test/java/com/mars_sim/core/person/ai/CacheCreatorTest.java
+++ b/mars-sim-core/src/test/java/com/mars_sim/core/person/ai/CacheCreatorTest.java
@@ -1,13 +1,88 @@
 package com.mars_sim.core.person.ai;
 
 import com.mars_sim.core.data.Rating;
+import com.mars_sim.core.data.RatingScore;
 import com.mars_sim.core.time.MarsTime;
+import com.mars_sim.core.tool.RandomUtil;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.mockito.InjectMocks;
 import org.mockito.Mock;
+import org.mockito.MockedStatic;
+import org.mockito.MockitoAnnotations;
+
+import java.util.ArrayList;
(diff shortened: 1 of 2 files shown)


mokun:
@muybien3433 

If you do find it is necessary to rename commonly used methods such as *setContainerUnit()*, *isInGarage()*, *transfer()*, etc. to improve code clarity, do coordinate a little bit with us first. This will help deconflicting/minimizing bad merges if we happen to be working on certain class that are utilizing them.

@bevans2000 

You're working on improving [Unit Hierarchy](https://github.com/mars-sim/mars-sim/issues/1197). Can you comment which class you don't want us to touch so that it won't make your PR unsuccessful ?

mars-sim-core/src/main/java/com/mars_sim/core/vehicle/Vehicle.java
```diff
@@ -1776,109 +1773,85 @@ public SalvageInfo getSalvageInfo() {
...
 			}
 		}
 
 		if (foundGoodLocation) {
 			setParkedLocation(newLoc, newFacing);
-			logger.info(this, "Moved to new parking loc at (" 
-					+ Math.round(newLoc.getX() * 10.0)/10.0 
-					+ ", " + Math.round(newLoc.getY() * 10.0)/10.0 
+			logger.info(this, "Moved to new parking loc at ("
+					+ Math.round(newLoc.getX() * maxAttemptsPerRange)/maxAttemptsPerRange
```
mokun: Math.round(abc * 10.0)/10.0 is a common way of deriving a double with just 1 decimal place.

Math.round(abc * 10)/10 is not the same as Math.round(abc * 10)/10. 

In this situation, the effect is not severe since it's only printing the result of a new x and y coordinate.

Just so you know.

bevans2000:
Yes I'm making changes in this area. Can we split this PR to just cover the Cache creator changes? Can you cherry pick the last commit for the Cache creator into a new branch and PR that?
I don't want to change the transfer method name because this is going to be 'enshrined' in an interface because we need to make this all generic.
The LocationStateType is a good example of what I'm trying to achieve. Unit has a lot of methods that are not applicable for most of the Units because only a subset can actually move around

bevans2000:
Should have said, my goal is to have something to push by weekend.

mokun:
> The LocationStateType is a good example of what I'm trying to achieve

Are you planning to get rid of the use of LocationStateType ?

mokun:
> I don't want to change the transfer method name because this is going to be 'enshrined' in an interface because we need to make this all generic.

Right. You will implement that change we've discussed in [here](https://github.com/mars-sim/mars-sim/issues/1356).

@muybien3433

Guess you can revert this [one](https://github.com/mars-sim/mars-sim/pull/1436/commits/e11ac320624f0ab5077b8d039651eb050529f67b), 

Renaming transfer() to transferVehicle() is kind of breaking a pattern we've adopted in the transfer() method used in all 4 class, namely, Vehicle, Person, Robot, and Equipment, etc.

mokun:
@muybien3433 

It looks like you revert the transfer() method name change in [here](https://github.com/mars-sim/mars-sim/pull/1436/commits/61bbe1f6735cc529eaadd173f3c68d4dd905bee5). Is it a full reversal of all instances of transfer() ? 

But somehow, your IDE is upload a bunch of groovy jar files into the *mars-sim-ui/lib" folder. 

Is that by accident ? 

See pic below: 
image

Do we need groovy at all in the mars-sim-ui package ? 

Those are jar files and are not supposed to be mixed with our source code repo.

image

If we do need groovy (I wonder why ?), we can define it as a maven artifact in pom.xml.

bevans2000:
> > The LocationStateType is a good example of what I'm trying to achieve
> 
> Are you planning to get rid of the use of LocationStateType ?

Not yet but it is only needed for Mobile units are currently everything has to support it. So eventually we can change how it works.
At the moment I want to minimise changes in this area this week.

mokun:
@muybien3433,
 
I just filed a new [issue](https://github.com/mars-sim/mars-sim/issues/1438) ticket on the CacheCreator. 

Let's revert some changes in this PR and be done with it.

Can you make a simple PR that will fix that NPE issue and merge it with the Master branch right away ?

Issue #1438, opened 2024-09-18
NPE on missionProbCache


mokun:
**Describe the bug**
- At Sol 2:000, missionProbCache becomes null.

**Expected and unexpected behavior**
- A person cannot get a new mission when calling MissionManager's getNewMission() if missionProbCache is null.

**Discuss any potential problems you have and have not ruled out**
- This is due to the use of the new CacheCreator generic class in a recent [PR](https://github.com/mars-sim/mars-sim/pull/1428)

**Java classes involved**
- MissionManager, CacheCreator.

In MissionManager, missionProbCache can become null from time to time
```java

	private MissionRating selectMissionFromProbabilities() {
		return missionProbCache.getRandomSelection();
	}
```

**Stacktrace**
```
01-Adir-02:000.445 (Info) MissionManager : [Terminus - Lander Hab 2] Joshua Young - Regolith Prospecting Score: 0.337 (base: 1, minerals: 0.805, leader: 0.31, extrovert: 1, settlement.ratio: 2.7, goods: 0.5)
01-Adir-02:000.445 (Severe) UnitManager : Problem with pulse on Terminus: Cannot invoke "com.mars_sim.core.person.ai.CacheCreator.getRandomSelection()" because "this.missionProbCache" is null
java.lang.NullPointerException: Cannot invoke "com.mars_sim.core.person.ai.CacheCreator.getRandomSelection()" because "this.missionProbCache" is null
	at com.mars_sim.core.person.ai.mission.MissionManager.selectMissionFromProbabilities(MissionManager.java:237)
	at com.mars_sim.core.person.ai.mission.MissionManager.getNewMission(MissionManager.java:221)
	at com.mars_sim.core.person.ai.Mind.startNewMission(Mind.java:439)
	at com.mars_sim.core.person.ai.task.PlanMission.selectingPhase(PlanMission.java:161)
	at com.mars_sim.core.person.ai.task.PlanMission.performMappedPhase(PlanMission.java:114)
	at com.mars_sim.core.person.ai.task.util.Task.executeMappedPhase(Task.java:680)
	at com.mars_sim.core.person.ai.task.util.Task.performTask(Task.java:640)
	at com.mars_sim.core.person.ai.task.util.TaskManager.executeTask(TaskManager.java:498)
	at com.mars_sim.core.person.ai.Mind.takeAction(Mind.java:201)
	at com.mars_sim.core.person.ai.Mind.moderateTime(Mind.java:173)
[...]

PR #1439, opened 2024-09-18
Edit MissionManager to resolve issue

muybien3433:
I believe that when CacheCreator was initialized as null  getRandomSelection method  couldn't satisfy condition 
"if (createdTime != null)" because createdTime was also null so it returned lastEntity(with is null) when (randomDouble <= probWeight) condition was true.

Assuming I created MarsTime instance right it should work. Due to #1438

### DIFF
--- a/mars-sim-core/src/main/java/com/mars_sim/core/person/ai/CacheCreator.java
+++ b/mars-sim-core/src/main/java/com/mars_sim/core/person/ai/CacheCreator.java
@@ -60,7 +60,7 @@ public class CacheCreator<T extends Rating> {
         T lastEntry = null;
         double randomDouble = getRandomDoubleBasedOnProbability();
 
-        for (T entry: cache) {
+        for (T entry : cache) {
             double probWeight = entry.getScore().getScore();
             if (randomDouble <= probWeight) {
                 if (createdTime != null) {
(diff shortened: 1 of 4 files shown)


bevans2000:
Sounds a reasonable assumption. You can't really create a MarsTime instance. We get them off the Master clock.

bevans2000:
Is this ready to merge to fix the problem?

mokun:
Right ! Besides the null MarsTime, I can see a typo mistake. 

In the old MissionManager,
> private transient CacheCreator<MissionRating> missionProbCache = null;

In the old getNewMission(),
> List<MissionRating> missionProbCache = new ArrayList<>();

Both are called missionProbCache.

muybien3433:
If you think 'MarsTime marsTime = Simulation.instance().getMasterClock().getMarsTime();' is right way to fetch, then it's ready to merge.

mokun:
> If you think 'MarsTime marsTime = Simulation.instance().getMasterClock().getMarsTime();' is right way to fetch, then it's ready to merge.

@bevans2000 

Would you prefer to bring in a static MasterClock instance here in MissionManager ?

But in doing so, it would mean that we must override this MasterClock instance with a new one when deserializing or loading from a saved sim and it's really a hassle with having to keep track of so many of these static instances already.

bevans2000:
Leave it as it is for now and we can sort this out later

PR #1444, opened 2024-09-20
Remove duplicates from BuildingManager

muybien3433:
Add more readability to this class.


mars-sim-core/src/main/java/com/mars_sim/core/structure/building/BuildingManager.java
```diff
@@ -2031,33 +1969,39 @@ public double getTotalEntropyPerLab() {
...
 					.stream()
 					.filter(b -> b.getZone() == person.getBuildingLocation().getZone()
 							&& !b.getMalfunctionManager().hasMalfunction())
 					.collect(Collectors.toSet());
 		}
 		else {
-			bldgs = bldgs
+			buildings = buildings
 					.stream()
 					.filter(b -> b.getZone() == 0
```
mokun, line 1966: As a sidenote, we haven't really applied the concept of district division or zoning much for building management yet.

But we'll revisit the need of zoning as the settlement grows to over a hundred buildings.

PR #1448, opened 2024-09-22
Improve EquipmentFactory

bevans2000:
Rework the ```EquipmentFactory``` class to improve the code quality.

The logic is fragile because it depends upon certain Manufacturing Processes that are found by name. This may break if the name of the process is changed or when internationalise of configurations is applied. To protect against this there is a new UnitTest that flags if the mappings are broken.

### DIFF
--- a/mars-sim-core/src/main/java/com/mars_sim/core/equipment/EquipmentFactory.java
+++ b/mars-sim-core/src/main/java/com/mars_sim/core/equipment/EquipmentFactory.java
@@ -7,10 +7,10 @@
 
 package com.mars_sim.core.equipment;
 
-import java.util.HashMap;
+import java.util.EnumMap;
 import java.util.Map;
+import java.util.Optional;
 
-import com.mars_sim.core.SimulationConfig;
 import com.mars_sim.core.UnitManager;
 import com.mars_sim.core.logging.SimLogger;
 import com.mars_sim.core.manufacture.ManufactureConfig;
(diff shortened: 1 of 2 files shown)


mars-sim-core/src/main/java/com/mars_sim/core/equipment/EquipmentFactory.java
```diff
@@ -142,39 +109,39 @@ public static double getEquipmentMass(EquipmentType type) {
...
-	    		if (info.getName().equalsIgnoreCase(processName)) {
-	    			manufactureProcessInfo = info;
-	    			break;
-		        }
-	    	}
+    private static double calculateMass(EquipmentType type) {
+		
+		// Note: it's haphazard to match the string of the manu process since it can change.
+		// Will need to implement a better way in matching and looking for the manu process that assemblies the item of interest.
+		String processName = switch(type) {
```
mokun, line 116: So we may as well find a way to look for and flag these manufacturing processes right away when processing ManufactureConfig at the start of the sim. 

They will be `ManufactureProcessInfo` instances.

If we can create a class for flagging these processes, we can use it for those haphazard process name strings in ItemResourceUtil as well.

bevans2000: Possible yes. Key point it works and the new unit test will tell us when it stops working.
It's part of the bigger question of how to handle locale in the configurations, but that's not that important

PR #1449, opened 2024-09-22
Cosmetic stuff in Vehicle

muybien3433:
Adds more clarity to class.


mars-sim-core/src/main/java/com/mars_sim/core/mission/MissionStep.java
```diff
@@ -99,28 +98,28 @@ protected void addLifeSupportResource(int crew, double durationMSol, boolean ide
      * @param task Task allocated
      */
     protected boolean assignTask(Worker worker, Task task) {
```
mokun, line 100: This assignTask() method is not being used at this moment.

The two real assignTask() methods (one for Person, one for Robot) that are in use are in AbstractMission. They are more refined than the old one here.

@bevans2000 
As a side note, personally speaking, I don't like the approach of first creating an instance of a task and then inserting it here as a parameter into assignTask() method. 

Many CPU cycles would have been wasted if the precondition for this task is not met and we would have to cancel this task.

mokun:
@muybien3433 

MissionStep and other associated classes inside this folder "mars-sim-core/src/main/java/com/mars_sim/core/mission/" aren't being put in use yet. Just so you know.

They are in work under this one big [project](https://github.com/orgs/mars-sim/projects/4).

mars-sim-core/src/main/java/com/mars_sim/core/mission/MissionStep.java
muybien3433: MissionDisembarkStep does

bevans2000: Sure but tat is a massive redesign of how Tasks work. Lets focus on achievable changes.

mars-sim-core/src/main/java/com/mars_sim/core/mission/MissionTravelStep.java
```diff
@@ -76,9 +76,7 @@ protected boolean execute(Worker worker) {
         boolean workedOn = false;
         if (vehicle.getOperator() == null) {
             Task operateVehicleTask = createOperateVehicleTask(vehicle, worker);
-            if (operateVehicleTask != null) {
```
bevans2000, line 79: This check is needed because the Mission cannot always assign Tasks. So sometimes there is no Task associated with the mission that can be started.

bevans2000:
@muybien3433 great you are contributing but be careful just refactoring the code because there may be active Issues being worked on.
Why not pick one of the unassigned issues off the backlog; maybe something off milestone 3.9.0

mars-sim-core/src/main/java/com/mars_sim/core/mission/MissionTravelStep.java
mokun: Perhaps it's okay here since createOperateVehicleTask() currently doesn't return null. 

It will either return a new instance of DriveGroundVehicle or throw a illegalArgumentException. 

I still don't like that we already start a DriveGroundVehicle task first.

Why can't we just test if it's assign-able first ? 

If it's not possible for this person to take on this task, we shouldn't even waste CPU cycles to create this task instance in the first place ?

Let's start a new ticket to address this.

The assignTask() method is being used in dozens of places. 

In the past, I've refactored the use of assignTask() for walking tasks.

Are there ways to make unit tests to test on various scenarios to see how much cpu utilization is wasted or saved by changing how it's being used ?

muybien3433:
> @muybien3433 great you are contributing but be careful just refactoring the code because there may be active Issues being worked on. Why not pick one of the unassigned issues off the backlog; maybe something off milestone 3.9.0

I’ll see what I can do. My point in refactoring, cleaning code is to better understand it.

bevans2000:
Totally agree, love a bit of cleaning code

Issue #1290, opened 2024-04-27
Create a seperate Settlement Template Configuration class


bevans2000:
**Is your feature request related to a problem? Please describe.**
SettlementConfig class play 3 roles:

1. Global configurations for Settlement, e.g. Shifts, Group Activities
2. Repository for SettlementTemplates using the existing UserConfiguration pattern
3. Holds FunctionSpec extra properties, e.g. ScienceLevel for a RESEARCH function.

This make the class very big and has a mixture of logic. Plus SettlementTemplates are transient and only needed when creating Settlements.

**Describe the goal(s) you have in mind**
Separate the Settlement template logic out into a new class called ```SettlementTemplateConfg```. This will focus solely on loading and saving Settlement Templates. 
In addition it will ensure that any FunctionSpec properties are solely within the template and not in the main SettlementConfiguration. The existing FunctionSpec will be subclassed for the Function that need extra configuration properties.

mokun:
> Separate the Settlement template logic out into a new class called SettlementTemplateConfg. This will focus solely on loading and saving Settlement Templates.

Sure. Good idea to make it simple.

bevans2000:
@muybien3433 this is a good one to understand how the configurations work

PR #1452, opened 2024-09-23
Create SettlementTemplateConfig class

muybien3433:
I created a new class to separate the templates from settlements config as #1290 says.

This required updating all instances of getItem() because UserConfigurableConfig was moved from SettlementConfig to the new class.

You can marge if you believe it's correct.

### DIFF
--- a/mars-sim-core/src/main/java/com/mars_sim/core/SimulationConfig.java
+++ b/mars-sim-core/src/main/java/com/mars_sim/core/SimulationConfig.java
@@ -13,6 +13,7 @@ import java.time.format.DateTimeFormatter;
 
 import javax.xml.XMLConstants;
 
+import com.mars_sim.core.structure.SettlementTemplateConfig;
 import org.apache.commons.io.FileUtils;
 import org.jdom2.Document;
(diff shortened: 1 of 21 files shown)


mokun:
Wonderful !

mokun:
Wonderful !

@bevans2000 

Shall we merge ?

mars-sim-core/src/main/java/com/mars_sim/core/SimulationConfig.java
```diff
@@ -268,7 +270,7 @@ private String loadValue(Element parentConfig, String child) {
 		String str = childItem.getAttributeValue(VALUE);
 
 
-		if ((str == null) || str.trim().length() == 0)
+		if ((str == null) || str.trim().isEmpty())
```
mokun, line 273: we can even use [isBlank()](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/String.html#isBlank()) method in future in cases like this as I was told that it's a bit more efficient than .trim().isEmpty().

> if ((str == null) || str.isblank())

Issue #1385, opened 2024-08-27
Cache the a map of coordinates to speed up mineral layer rendering


mokun:
**Is your feature request related to a problem? Please describe.**
- Currently, if the mineral layer is enable in Mars Navigator, the rendering of each color dot superimposed on a surface map are very *expensive* and unnecessarily repetitive.

**Describe the rationale**
- Get rid of the need of repeatedly calling `IntegerMapData.convertRectToSpherical()` 

**Describe the goal(s) you have in mind**
- Create a cache mechanism and combine x, y and rho into an index as the **key** of a map. This map will have Coordinates as the **value**.

**Describe the solution you desire to see**
- In Coordinates class, we have
```java
	/**
	 * Converts linear rectangular XY position change to spherical coordinates with
	 * rho value for map.
	 *
	 * @param x              change in x value (# of pixels or km)
	 * @param y              change in y value (# of pixels or km)
	 * @param rho            radius (in km) or map box height divided by pi (# of pixels)
	 */
	public Coordinates convertRectToSpherical(double x, double y, double rho) {
		Point2D point = IntegerMapData.convertRectToSpherical(x, y, phi, theta, rho);
		return new Coordinates(point.getX(), point.getY());
	}

```

In MineralMapLayer, we have a double loop structure : 
```java
	/**
	 * Displays the layer on the map image.
	 * 
	 * @param mapCenter the location of the center of the map.
	 * @param baseMap   the type of map.
	 * @param g         graphics context of the map display.
	 */
	public void displayLayer(Coordinates mapCenter, Map baseMap, Graphics g) {
		
...
			for (int x = 0; x < Map.MAP_BOX_WIDTH; x = x + 2) {
				
				for (int y = 0; y < Map.MAP_BOX_HEIGHT; y = y + 2) {
						
					// Do NOT delete the following line of codes
//					if (mapCenterCache == null || rho != rhoCache || numMineralsCache != numMinerals) {
					
						java.util.Map<String, Double> mineralConcentrations = 
							mineralMap.getSomeMineralConcentrations(
										mineralsDisplaySet, 
										mapCenter.convertRectToSpherical(x - centerX, y - centerY, rho), 
										mag);
						hasMinerals = true;

```

In RandomMineralMap, the method below gets called in by that double loop. It create the fuzzy look of each mineral concentration. It has 2 while loops inside and is thus very repetitive and needs to be cached or simplified.

```java 
	/**
[...]

Issue #946, opened 2023-06-23
Make use of 30 cartographic quadrangles for tile loading of surface maps


mokun:
**Is your feature request related to a problem? Please describe.**
- Enable tile loading of surface maps by considering well-defined regions called *cartographic quadrangles* on Mars.

**Describe the rationale**
- This saves us from having to define our own convention and roll our own implementation when there's one out there already.
- It makes it easy down the road if we switch to adopting the map and datasets what they will publish. 

**Describe the solution you desire to see**
- Make use of the United States Geological Survey (USGS)'s definition of the 30 cartographic quadrangles (MC-1 to MC-30) for defining the surface maps.

**Screenshots**
![image](https://github.com/mars-sim/mars-sim/assets/1168584/b57fd346-6513-44af-b45d-c94abfaf2e36)

**Specifications  (please complete the following information):**
 - pre 3.6.0 build 8271

**Additional context**
- Efficient method of map loadings or tile loading have been mentioned here and there in [#867](#867), [#868](#868), 
-  [#895](#895), [#901](#901),  [#916](#916),  [#924](#924) and [#935](#935). The definition of these regions of map may help.

mokun:
This illustrates the idea of tile loading of a big surface map: 
![deepzoom](https://github.com/user-attachments/assets/1d9cc510-7dc1-4f4f-ac7d-69adc283aa44)

bevans2000:
@mokun Are these "tiles" layered? I think they are just a single layer with no zooming possible.
Are they overlapping? Could we convert them into a set of rectangles? The shape makes them difficult to work with.
If we could do them as a set of Rectangle I think I can do it. For example, if we took our largest single image and separated it into a number of small images using a grid system then I could get something working.
I've had the idea when thinking about #1385

mokun:
I suppose they are all raw rectangular images. 

But for those mear the top or bottom, they will need to be 'projected' or  'stretched' correctly. 

We can pre-process or chop up one of our res level 4 map to try out how to do tile loading.

Do you think we should also consider how to load a small region of a map fragment together while we are working on tile loading ?

Both are almost the two sides of the same coin.

bevans2000:
The key is we have everything hidden behind the MapData interface which we are nearly there, making no assumptions about the underlying storage.

bevans2000:
One challenge I see is the MapData.createImage method is not async. So with a single image, the Map data can be in a loading state, once loaded create image can work.
[...]

Issue #916, opened 2023-06-15
Create super resolution elevation map around known destinations


mokun:
**Is your feature request related to a problem? Please describe.**
- Elevation affects movement of a vehicle. When elevation data is of low resolution, there's quite a bit of guesswork in calculating elevation between two given points.  

**Describe the solution you desire to see**
- Create a separate elevation data double array that will encircle around known destinations such as settlements.
- The elevation data will be super high resolution, say, one data point per meter.
- They will be pseudo-randomly generated upon the start of a settlement.
- The height boundary of this double array on all four sides should be seamless with the actual MEDGR data which is also a elevation double array with global data.

**Describe the rationale**
- When returning a vehicle from mission to a well-known, well explored destination such as a settlement, settler should know exactly which route to take to get home at fastest and most efficient manner

**Specifications  (please complete the following information):**
 - pre 3.6.0 build 8180

**Additional context**
- This is somewhat related to #901, where the ability of choosing which resolution of elevation data matters.

bevans2000:
I still believe that if we implement loading the high resolution map on the fly for a region around the target we can also achieve this purpose. 
Also we need to consolidate the various tickets around this area.

mokun:
One problem with that high resolution map is that The map scale is only 3.705 km per pixel.

If we ever want road network or want to draw a 2.5D settlement, then we need a 1000x better elevation map.

So we pretty much need to procedurally generate one of we want to, say, having the player to do a  first person EVA walk right outside the settlement.

But for now, I could settle with, say,  just 5 m per pixel randomly generated elevation data for vehicle driving.

So, of we are to consider elevation when placing a foundation and constructing a building, no data set currently available from NASA will suffice.

mokun:
>  implement loading the high resolution map on the fly for a region around the target 

Yes I like to implement this idea as well, in conjunction with the approach above. 

MEDGR website has even higher resolution data set but they are by regions.

We can modify TerrrainElevation class to try to read those into mars-sim. 

However, the regional area is still too big.

Even better is if we find a way to partition those big MEDGR img files into smaller regional chunks than the way they published them.

mokun:
@bevans2000 

The current elevation caching mechanism uses *index* for reading elevation data.

What does it take to modify it to support "artificial" elevations ? 

We currently bundle `megt90n000eb.img`. The elevation data has not good enough for piloting vehicle. 

[...]

PR #1455, opened 2024-09-28
Implement asynchronous map loading

### LINKED ISSUE #1037: Download new maps asynchronously
- If running mars-sim for the first time in a new machine, choose a map with resolution higher than level 0 will likely require downloading maps from github. 
- Since the size of the highest resolution of maps can be over 130 MB, Mars Navigator shows a black box and is being uninformative on what's going on. 

- Allow downloading a map from the internet asynchronously.
- Have a top level modal window pop up, pause the simulation, and provide a progress bar to show percentage of download.
- Allow the newly downloaded map to be loaded afterward.

**Describe the rationale**
- This gives player a better user experience.

 - pre 3.7.0 build 8634
 
Note: see #901 regarding maps

bevans2000:
This PR delivers the asynchronous loading of map for the Navigator tool. This has required a number of changes, including at a high level:

- ```FileLocator``` class can download remote file asynchronously using a background executor. Callers are notified via a ```Consumer```
- Map logic related to the UI are moved to the UI module. Keep change is ```Map``` becomes ```MapDisplay``` to remove the name clash with the standard Java Map.
- ```MapDisplay``` takes over from ```MapData``` in holding the current zooming factor. This means MapData instances become immutable once loaded.
- Slight change to the Zoom slider logic to use MapData as the source of truth of the zooming level. This removes the problem of synchronisation as background maps are loaded. Zooming limits are driven by the min & max RHO of the MapData.
- Navigator UI updated so map levels are shown as drag right menus; already loaded map layers are decorated.

There is also an optional UI feature to show details of the map data controlled by ```MapPanel.SHOW_MAP_DETAILS``` property.

close #1037


bevans2000:
@mokun I've put together a draft PR so you can see what I've done to the mapping. I think it is a lot cleaner and easier to follow.
I've got some further cleaning/refactoring but that wont affect the basic behaviour.
Can you have a go with it. The zooming works like 3.8, I think. But I'm not sure its make sense.

mokun:
>  Zooming is driven by the min & max scale of the selected map which I think is totally correct, but I think the way we calculate min & max needs to change.

So after the change, what's your definition of scale ? 

Ideally speaking, we could define scale as the true magnification ratio such as something like 1:20,000 at a certain zoom level.

On google map, when I open it, it shows ~1 true inch bar on the bottom right that equals 1000 virtual ft (or under metric system, say ~1 true cm equals to 200 virtual meter).

bevans2000:
I have tried to replicate what was there previously

mokun:
So does it mean that the copy you are working off at is the latest one form the master branch ?

bevans2000:
You can see which branch it is from the pull request. It has your zoom changes but the recent changes are not there. It won't make a materialistic difference to the zooming.
Are you going to give it a try?

mokun:
No worry. I don't mind you go ahead with the merge whenever you considered it done.

After that we'll refine whatever features/improvements we see fit.

mokun:
> Scaling setting held in UI layer and not MapData

Actually all the work regarding the map won't be needed by the console version at this moment.

mokun:
It will be great one day we can tap the use of topo maps to create better, more natural mission trails that are conforming to topographical/terrain changes. 

Only in this case will we need to load maps for the console version.

Will we tackle on this issue of straight line mission trail soon by any chance ?

bevans2000:
@mokun there is one significantly problem left affecting #946 relating to the ```Integer::getMapBoxPoint``` method. This uses a singularly loaded static array. This method is tightly coupled with the loading of a single image but we need everything to go via the ```MapData``` interface.
The underlying root cause is how the mineral map layer is drawn. We need to resolve #1385 before we can do the tiling.

mars-sim-core/src/main/java/com/mars_sim/core/environment/TerrainElevation.java
```diff
@@ -287,233 +263,6 @@ public double obtainRegolithCollectionRate(Coordinates loc) {
...
-	
-	/**
-	 * Compute the HSB color topo map based Elevation.
-	 * 
-	 * @param phi
-	 * @param theta
- 	 * @param rho
-	 * @return elevation in km
-	 */
-	public static double getColorElevation(double phi, double theta) {
```
mokun: I just added the capability back in this [commit](https://github.com/mars-sim/mars-sim/commit/55e8bef6d3c858177c0e8c797bb8f97b82b88490) recently.

Topo map by definition conveys the topographically info to the viewer. 

Thus, it connects terrain to an elevation range. 

This method will eventually supplement the elevation data encoded as  a pixel color. 

However, because of jpg compression, I found that the recovered RGB colors has been degraded and no longer matches up. Therefore, the HSB values are off. I bet the tiff image (or the uncompressed jpg image) would give better result.

I will come back to this in future.

mars-sim-core/src/main/java/com/mars_sim/core/environment/TerrainElevation.java
```diff
@@ -287,233 +263,6 @@ public double obtainRegolithCollectionRate(Coordinates loc) {
...
-//			if (Math.abs(location.getPhi() - SOUTH_POLE_PHI) < .04D) {
-////				if (elevation < 2D)
-//					result = .783;
-//			}
-//		}
-          
-    	return result;
-    }
-    
-	public int getHue(int red, int green, int blue) {
```
mokun, line 489: Note that the Hue (H) from HSB value encodes the height information under the DEM/DTM model. 

That's why we are experimenting this approach here as explained in this [thread](https://github.com/mars-sim/mars-sim/issues/916).

mars-sim-core/src/main/java/com/mars_sim/core/map/IntegerMapData.java
```diff
@@ -66,10 +67,8 @@ public class IntegerMapData implements MapData {
 	private int pixelHeight;
 	/* The cache for the last rho */
 	private double rhoCache;
-	/* The cache for the last phi. */
-	private double centerPhiCache;
-	/* The cache for the last theta. */
-	private double centerThetaCache;
+	/* The cache for the last center. */
+	private Coordinates centerCache;
```
mokun, line 71: It's more efficient to simply pass 2 primitive data members namely, phi and theta) than the Coordinates object.

Both the regular cpu0() or the openCL's gpu() methods needs to processs the raw phi and theta.

Inside cpu0(), it calls for convertRectIntToSpherical(). Inside it needs the value of phi and theta.

I made this change recently to speed up the mineral layer rendering. 

Getting rid of using coordinates help with the efficiency tremendously.

mars-sim-core/src/main/java/com/mars_sim/core/environment/TerrainElevation.java
bevans2000: The point is the colour terrain lookup is only used by the Console code. So we shouldn't be keeping code that is not used

mars-sim-core/src/main/java/com/mars_sim/core/map/IntegerMapData.java
bevans2000: Ok,but I know that using the Coordinate is the interface will have no difference on performance. It's about consistency in design and simple interfaces. The Coordinate class just wraps the 2 double, good design should separate them only when we need in the internal method like cpu

mars-sim-core/src/main/java/com/mars_sim/core/environment/TerrainElevation.java
mokun: >  colour terrain lookup is only used by the Console code.

Yeah that it went through changes lately not to bother using it since more work needs to be done to parse the color of a pixel into useful elevation data.

Issue #1388, opened 2024-08-28
Add new tools for Mars Navigator


mokun:
**Is your feature request related to a problem? Please describe.**
- When it comes to mission planning for expeditions, Mars Navigator should be the go-to toolbox for players.
- Currently, this tool only has a zoom slider, a button for selecting a surface map and the option of turning on and off a myriads of map layers. 
- We propose the following tools to be made available 
   - A ruler for measuring distances between any two points 
   - A tracer tool for creating topo lines (not straight lines)
   - A area measurement (square km) tool for estimating the the mineral reserve volume or mass  
- Some of the capabilities would be shared with the surface map in Mission Tool's Nav tab, which is also based on Mars Navigator.

**Describe the rationale**
- Mars Navigator can be more feature rich for visual mission planning. 

**Describe the initial goal(s) you have in mind**
- Drop custom markers
- Trace topo lines between two markers
- Measuring distances (along a selected topo line or a straight line) between any two markers 

**Describe the non-goal(s) associated with the goal(s)**
- In future, rework the switching of surface map's resolution level to be more player-friendly
- Allow two maps to be simultaneously loaded side-by-side for visual comparison

**Describe the solution you desire to see**
- Add mouse controls for defining a start point and an end point
- Add algorithms for tracing topo lines (lowest possible change of adjacent color pixels) 

**Specifications**
- post-3.8.0

mokun:
@bevans2000 

Btw, I found out not all maps are centered (say at Latitude 0 and longitude 0) at the same location. They can be a little bit off.

Thus a settlement may be shown inside a small crater in one surface map but outside of that same crater on another surface map. 

I may need to pull back the newly added *Vertical Roughness* map.

mokun:
@bevans2000 @muybien3433 

I'm relocating the zoom slider bar that's on the right hand side of Mars Navigator.

It has been attached to Mars Navigator. I'll make it float on top of the map and on the right hand side of the map, much like how it appears on the Settlement Map. 

This way, the nav tab map may be given the capability of the zoom slider as well.

bevans2000:
@mokun are you finished with you changes to the Map Tool for a short period? I want to continue with #1037 but will now need to make changes to the Map tool itself

mokun:
> are you finished with you changes to the Map Tool for a short period? 

Sure. I'll pause.

bevans2000:
> > are you finished with you changes to the Map Tool for a short period?
> 
[...]

PR #1458, opened 2024-10-05
Remove the displayUnit concept from UnitManager

### LINKED ISSUE #1408: Remove the displayUnits from UnitManager
**Describe the bug**
The display unit list is an artificial list that contains all Vehicles and Settlements. The is only used by the Navigation tool and that is where the functionality should reside' not in the generic UnitManager.

**Change**
Remove the displayUnit from UnitManager.
Rework Navigation Panel to directly use Settlement and Vehicles lists from the Unit Manager.

bevans2000:
This removes the displayUnit concept from the ```UnitManager```. This is a UI concept and should not be in the core module.
The specific layers that use display units now explicitly render Settlements and Vehicles.

A new ```MapMouseListener```class  is created that centralises the mouse selection logic in the Navigator window and NavPointPanel. 

close #1408

### DIFF
--- a/mars-sim-core/src/main/java/com/mars_sim/core/UnitManager.java
+++ b/mars-sim-core/src/main/java/com/mars_sim/core/UnitManager.java
@@ -140,7 +140,7 @@ public class UnitManager implements Serializable, Temporal {
 	 * @return
 	 */
 	private Map<Integer, ? extends Unit> getUnitMap(UnitType type) {
-		Map<Integer, ? extends Unit> map = switch (type) {
+		return switch (type) {
 			case PERSON -> lookupPerson;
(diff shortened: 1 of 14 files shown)


mokun:
@bevans2000 

In conjunction with this PR, would you be interested in adding a color palette tool in Mars Navigator ? 

This way, player can freely customize and choose what color they want to show the settlement/landmark/vehicle names on a particular surface map. 

It's hard to pre-determine what color scheme is needed for labels since each color/grey map has a different background color tone.

See https://github.com/mars-sim/mars-sim/issues/1388

bevans2000:
I like that idea. I have got another draft PR pending that cleans up a few things. 
I want to at least improve the mineral layer rendering

PR #1460, opened 2024-10-14
Rework MapLayer logic

bevans2000:
This PR focuses on refactoring the MapLayer logic in the Navigator Window as part of #1385. It addresses the following high level changes:
- Creation of the concept of a Surface Point of Interest based at a Coordinate that is well aligned to entities rendered on a Map layer
- A ```SurfaceManager``` class that provides a means to store ```SurfacePOI``` and provide efficient  geometric based searching
- A common ```SurfaceFeatureMapLayer``` to consolidate logic on selecting and rendering ```SurfacePOI```

The above are applied to Landmarks and Explore Sites.

A further optimization is to significantly reducing the searching required for tracking objects on the map. This is done by the map layers reporting what hotspots are valid. This allows a single small list to be maintained of only the visible hotspots.  
In addition, the hotspots now supports a tooltip capability that is available to Landmarks & Explored Sites.

There are various clean-ups on ```MapLayer```.


mars-sim-core/src/main/java/com/mars_sim/core/structure/Settlement.java
```diff
@@ -917,54 +914,6 @@ public boolean timePassing(ClockPulse pulse) {
 		return true;
 	}
 	
-	/**
-	 * Checks and reports on the average mineral site improvement made.
-	 * 
-	 * @param sol
-	 */
-	private void reportsMineralSiteImprovement(int sol) {
```
mokun, line 925: Can we still have this method for reporting mineral site improvement at the end of a sol ?

mars-sim-core/src/main/resources/messages.properties
```diff
@@ -914,6 +912,18 @@ NaturalAttributeType.teaching           = Teaching
 
 NavButtonDisplay.log.mediaTrackerError = NavButtonDisplay Media Tracker Error {0}
 
+LandmarkType.AO = Artificial Object
```
mokun, line 915: It'll be great that one day we offer players the option to turn on and off each of these landmark types, wouldn't it ?

bevans2000: Yes I agree. I want it that each layer can support filters. I'm working on ideas for it

mars-sim-core/src/main/java/com/mars_sim/core/structure/Settlement.java
bevans2000: It is an expensive method and the output just goes to the console. If you want to record the historical improvements of sites then lets do it probably and uses our Metrics Manager idea

mokun:
It seems after this PR or previous commits, the sun data panel no longer gets updated.

image

bevans2000:
I doubt it is this PR

PR #1469, opened 2024-10-26
Use Worker in TaskManager

### LINKED ISSUE #1466: Task manager should use Worker
The ```Taskmanager``` class uses a Unit as the target worker but not all Units can work on Tasks; only ```Worker``` objects. This inconsistency will create a future problem when the base Unit class is simplified and redundant logic removed.

**Describe the rationale**
This change will simplify the scope of the ```TaskManager``` and reduce the impact of future changes.

Change the ```TaskManager``` to use a ```Worker``` as the target. This involves changing the reinit method to take a worker when Person/Robot is restored.
In addition this will also stop the protected access to fields on the base class by forcing use of get methods.

bevans2000:
This PR covers changes to the Worker interface so it can be used in the abstract Task manager. This involves creating a MobileUnit interface that captures all movement methods.
The Task manager now is based on the Worker instead of Unit. This reduces the coupling to Unit.

close #1466

### DIFF
--- a/mars-sim-core/src/main/java/com/mars_sim/core/person/ai/task/util/Worker.java
+++ b/mars-sim-core/src/main/java/com/mars_sim/core/person/ai/task/util/Worker.java
@@ -7,22 +7,20 @@
 
 package com.mars_sim.core.person.ai.task.util;
 
+import com.mars_sim.core.UnitEventType;
 import com.mars_sim.core.UnitIdentifer;
 import com.mars_sim.core.UnitListener;
 import com.mars_sim.core.UnitType;
 import com.mars_sim.core.equipment.EquipmentOwner;
-import com.mars_sim.core.map.location.Coordinates;
-import com.mars_sim.core.map.location.LocalPosition;
 import com.mars_sim.core.person.ai.NaturalAttributeManager;
 import com.mars_sim.core.person.ai.SkillOwner;
 import com.mars_sim.core.person.ai.mission.Mission;
 import com.mars_sim.core.structure.Settlement;
-import com.mars_sim.core.structure.building.Building;
 import com.mars_sim.core.structure.building.function.ActivitySpot;
 import com.mars_sim.core.structure.building.function.ActivitySpot.AllocatedSpot;
-import com.mars_sim.core.vehicle.Vehicle;
+import com.mars_sim.core.unit.MobileUnit;
 
-public interface Worker extends UnitIdentifer, EquipmentOwner, SkillOwner {
+public interface Worker extends UnitIdentifer, EquipmentOwner, SkillOwner, MobileUnit {
 
 	/**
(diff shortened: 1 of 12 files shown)


mars-sim-core/src/main/java/com/mars_sim/core/person/ai/task/util/Worker.java
```diff
@@ -145,27 +92,6 @@ public interface Worker extends UnitIdentifer, EquipmentOwner, SkillOwner {
...
-	 * @param position
-	 */
-	public void setPosition(LocalPosition position);
-
-	/**
-	 * Sets the building the worker is located at.
-	 *
-	 * @param position
-	 */
-	public void setCurrentBuilding(Building building);
```
mokun, line 167: guess we no longer need this now for proactively setting its whereabout. 

Is it currently done by the activity spot related methods ?

mars-sim-core/src/main/java/com/mars_sim/core/unit/AbstractMobileUnit.java
```diff
@@ -0,0 +1,245 @@
...
+		return unitManager.getBuildingByID(currentBuildingInt);
+	}
+
+	/**
+	 * Computes the building the person is currently located at.
+	 * Returns null if outside of a settlement.
+	 *
+	 * @return building
+	 */
+	public void setCurrentBuilding(Building building) {
```
mokun, line 86: Never mind. Looks like we still need to call this method. It's simply being relocated to this class.

mars-sim-core/src/main/java/com/mars_sim/core/unit/AbstractMobileUnit.java
```diff
@@ -0,0 +1,245 @@
...
+
+    
+	/**
+	 * Gets vehicle person is in, null if person is not in vehicle.
+	 *
+	 * @return the person's vehicle
+	 */
+	@Override
+	public Vehicle getVehicle() {
+		if (getLocationStateType() == LocationStateType.INSIDE_VEHICLE) {
```
mokun, line 200: So, will you be removing the need of calling LocationStateType next and strictly using container unit for location determination ?

bevans2000: Not sure yet, it would be good if we could but taking this slowly. I have created a new project to cover all related chabges

mars-sim-core/src/main/java/com/mars_sim/core/person/ai/task/util/Worker.java
bevans2000: Good question, but will still need it. The point these methods are now moved to the Mobile unit interface. A Worker is also mobile which makes perfect sense

Issue #1514, opened 2025-01-11
Improve Manufacturing Process


bevans2000:
**Is your feature request related to a problem? Please describe.**
As part of #1236 a Manufacturing manager has been introduced that consolidates the queuing of Salvage and Manufacturing. It also changes the MetaTask approach to follow the Settlement backlog manner.
However this solution is complicated by manufacturing and salvage processing being implemented by separate classes. Furthermore, these classes have a lot of duplicated code.
Secondly, only 3d printers are available but some processes need other tools. E.g. baking bricks needs a kiln.

**Describe the goal(s) you have in mind**
Simplify the 2 sets of processing by having a single framework that supports activities in a workshop.

**Describe the solution you desire to see**
The proposal is to create a new Task that replaces the ManufactureGood and SalvageGood called ```WorkshopTask```. This is a new Task that moves a process through the manual working time logic through to completion. This new task would work on a ```WorkshopProcess``` that contains all the common attributes of the SalvageProcess and ManufactureProcess. The latter classes become subclasses.

Furthermore, these subclasses would take on the responsibility of preparing and completing the process which is currently in the Manufacture function. The object is the Manufacture Function just becomes a facilitator of the processing being responsible for advancing the processing time of the process.

The Workshop processes would have to implement a start and complete methods as well as abort. This would cover the different process to produce the correct goods; and also maybe future processes.

The benefits is there is no difference between the running on the 2 different process types and hence can simplify the Manufacture class as it just a common set of processes active. It also virtually halves the amount of code involved.
Having a single logic would make it easier to add extra facilities to a Manufacture function that could be utilised by different process info definitions

mokun:
Currently, it creates an excessive number of logs. I have a hard time scrolling back to see what I want to look at. The stacktrace buffer easily got filled up and cutting off the old logs.

bevans2000:
Does it? Thought it had little log. I can check.
Do you have an example?

bevans2000:
> Currently, it creates an excessive number of logs. I have a hard time scrolling back to see what I want to look at. The stacktrace buffer easily got filled up and cutting off the old logs.

You mean the current approach? I haven't made these changes yet

mokun:
I should have commented the excessive log here at https://github.com/mars-sim/mars-sim/pull/1515.

Issue #1236, opened 2024-01-17
Have one manufacturing queue for entire settlement 


mokun:
**Is your feature request related to a problem? Please describe.**
- Currently, there is no queue for manufacturing.
- When a lane or slot is available, the settlement will compute what process to undertake
- If there are more than one lander hab, workshop, manufacturing shed, machinary hab, players have to open all of them to discover what the settlement is manufacturing

**Describe the rationale**
- Player commander should be given better decision-making authority to decide what to manufacture. 

**Describe the goal(s) you have in mind**
- Create a Manufacturing tab in Monitor Tool to show what the settlement will manufacture
- Show the rating score on each process
- Allow manually adding and deleting manufacturing processes
- Add a manufacturing queue for storing future manufacturing processes
- Allow players to click to increase or decrease priority on a manufacturing process

**Specifications**
- post 3.7.0 build 9292

**Additional context**
- May implement these features in line with https://github.com/mars-sim/mars-sim/issues/1062

mokun:
@bevans2000 

In a lot of city building games, there's one central construction queue that shows what is being built.

We could in principle create one single queue panel that let player see all the manufacturing processes that are running among Workshops, Lander Hab, Machinery Hab, etc.

This allows player to decide if no overlapping effort in making the same part or resources among various buildings is warranted.

bevans2000:
Sure. I like this idea.

bevans2000:
What we need is a new class called ProductionManager that can maintain the queue and act as an overseeing for the various production functions. Then use a ScheduledEvent to periodically add to the production queue. Should be a nice new feature.
I'm struggling for time at the moment because of work and working on #1285 which is good because I am doing lots of new UnitTest for the Tasks.
Also drop the ability to creation new processes for individual Buildings in the UI but have it centralised. The metatask logic will be much easier as it just acts as a proxy to the ProductionManager for the Settlement being checked.

mokun:
For now, I'm going to simplify add the ability to queue in each building that has manufacturing function. 

So far, if a workshop supports only 2 concurrent processes and there are already 2 on-going processes. Player cannot add the 3rd one. 

This is very inconvenient since player has to wait for those 2 processes to get done first before he can add one. 

Also, if the settlement decides to add 2 processes right after the previous 2 was done. The players has to wait again.

Adding a queue now for each building makes sense.

In future, we'll consider a settlement-wide queue.

bevans2000:
[...]

Issue #941, opened 2023-06-21
Improve timekeeping by defining time zones on Mars


mokun:
**Is your feature request related to a problem? Please describe.**
- Human beings are used to define the beginning of the day as morning. The early hour is when the quantity of hours is small. As it progresses, the number gets bigger. At midnight, when it transitions onto the next day, the number goes to zero again.
- Having a time reference to a local activity is crucial. Say, I used to wake up at 8 am every morning and now I'm being asked to wake up one hour earlier (at 7 am) to go to school/work. 
- However, at 7 or 8 am is a time reference useful to me only. If my friend is on a Moon base and he wants to know exactly when I wake up to give me a call, it's not meaningful enough for him to know that I wake up at 7 or 8 am. He needs to know in what time zone I'm in and do the conversion so that he can call me when I'm awake.
- In other words, we not only want to keep track of time locally but also universally. We need both so that we can figure out what time we can meet remotely. 

**Describe the rationale**
- Now we are working our problem backward. We already have a Universal Martian Time Standard by default. What we don't have is a Local Time Standard. 
- In other words, we need a Time Zone system for settlers to conveniently keep track of time locally.
- The disadvantage of a local time zone is that as settlers scatter across Mars, they will no longer use one universal time standard. Transactions and other activities can become more difficult across the continent. It encourages independence.

**Describe the solution you desire to see**
- In [#743](https://github.com/mars-sim/mars-sim/issues/743), we discussed implementing a new shift system. We also mentioned the possibility of a Time Zone system.
- As @Urwumpe pointed out 
> why not a 24 time zones on an almost 24 h Mars?
- We could in fact have a 24 time zone. This [group](https://www.aresastro.com/2016/03/ares-time-zone-system/) sort of map out what a 24-hour time zone would look like on Mars.

```
Based upon the MTC standard, the Arean time zone system consists of 24 evenly spaced time zones, 
each 15° in longitude representing 1 hour intervals spanning from MTC-11 to MTC+12.
```

**Describe alternatives you've considered**
- Leave it the way it is. Stick to the universal time currently being used. 

**Specifications  (please complete the following information):**
- pre 3.6.0 build 8256

bevans2000:
This will need applying to the ```ShiftManager``` constructor as the Shift are offset by a time according to the rotation around the planet.

mokun:
[...]

PR #1515, opened 2025-01-12
Create Manufacturing Manger supporting a single queue per Settlement

bevans2000:
This PR delivers the majority of the #1236 requirements. 

- Creates a ```ManufacturingManager``` per Settlement that manages all manufacturing and salvage processes within a Settlement.
- This delivers a single queue of potential processes that are ordered according to the value of the outputs to the Settlement. This value is dynamic and changes to reflect the changing needs of the Settlement.
- A new design to the Manufacture panel in the Settlement unit window is provided that shows the Queued and Active separately. The former tab support the user adding new processes but this is more focussed that previous. Only processes that can be support according to skill & tech are offered. Selection is further simplified by the user firstly choosing the desired Output and then processes that can deliver that output.
- Manufacturing manager also automatically adds new Manufacturing processes to the queue under the user control of the user via the Preferences logic.

To support the change there has been improvement to modelling of Tooltips in tables. Creation of a time offset in Settlement for #941.

This is a work in progress PR with the remaining outstanding aspects are
- UI to all the user to change the weighting of queued processes and remove queued processes
- Automatically add Salvage processes
- Allow specific high level Units, e.g. Vehicle & Robots to be scrapped for this process.
- Whilst not directly related, the follow on issue #1514 

-


mokun:
Currently, it creates an excessive number of logs. I have a hard time scrolling back to see what I want to look at. The buffer of the stack trace window in eclipse got filled up.

bevans2000:
Ok I'll take a look but I don't remember adding any excessive log statements

PR #1581, opened 2025-06-07
Make the Navigator Window resizable

bevans2000:
The goal of this PR is to allow the Navigator Window to be resized by the user.
The first step was to remove the hard coded sizes of 512x512 in the MapPanel and all other classes. The solution is to drive the logic by the size of the MapPanel instead of the constants.
Second step was to allow the Navigator window to be resizable. The caching of the images is moved from the MapData classes into the MapDisplay classes. This because the MapData is shared by multiple MapPanels and hence provides no benefit.


bevans2000:
@mokun So redesigning the navigator means it is easier to make it sizable by the user. I've got it roughly working and should be able to push this tomorrow. This has long been an annoyance that the size of this tool is fixed.

mokun:
But in future, do we want to get rid of the concept of having different tools in an empty main window ? 

I would favor having a game-like environment, rather than window-based sets of tools. 

But I do understand both approaches have their own merits.

PR #1595, opened 2025-06-28
Create Mission Objectives

bevans2000:
This PR introduces the concept of a ```MissionObjective``` which records the outcome of a Mission. The purpose is to remove the direct dependency between the Mission and it's goals; the outcome is defined by the Phases that are part of the mission. This change will allow in the future Missions to have a mixed combination of Phases and hence multiple outcomes.
For example a Mission may do Exploration and resource collection the target locations are close to optimise any travel.

This change involves a reworking of the Mission tool to make the outcome panel be driven by the objectives of the viewed mission; not by the type of Mission. Eventually the type of Mission will become artificial.

As part of this reworking the new ```EntityLabel``` component is created. This provides an easy and consistent means to show an ```Entity``` in the UI. The benefit is this automatically provides button to show the appropriate Details window as well as the current location in either Map tool. A good example of the improvement is the presentation of the Scientific Study in the new Field Study panel; the user can now jump directly to the Study tool.

It is a partial change and further work is required to convert the remaining Mission classes.


mokun:
> A good example of the improvement is the presentation of the Scientific Study in the new Field Study panel; the user can now jump directly to the Study tool.

What if we can make each scientific topic an entity label ?

This way, we can track who or when a topic has been research. Perhaps we'll need a new UI panel to document and provide statistics.

I admit that, for now, scientific topic doesn't do anything yet.

But I can think of parts (items) or equipment that can be invented down the road once a scientific topic has been sufficiently studied.  

Or an existing part or equipment can be improved in terms of its performance specs.

Issue #1298, opened 2024-05-06
Restructure Package hierachy


bevans2000:
**Is your feature request related to a problem? Please describe.**
The source code has grown over time and the existing package structure results in hotspot of large packages. 
Secondly there are cases where the separation of concerns is poor giving a high dependency.

**Describe the rationale**
Split the packages into a more structured approach with more top-level packages.

**Describe the goal(s) you have in mind**
Key changes would be:

- ```structure``` renamed to ```settlement```
- ```structure/building``` moved to new top level package
- ```structure/construction``` moved to new top level package
- ```person/ai/task``` moved to new top level package

The Tasks classes will be moved out of the general single location into ```task``` sub-packages under the appropriate top level; this is being covered as part of #1285.

This would be a large change but relatively low risk if done in an IDE. It's important files undergo a Git rename to preserve the history.

mokun:
Can we also consider the possibility of having remote stations, such as weather tower, communication towers, traffic hub, etc. that will be minimally manned and/or robotically operated ?

bevans2000:
Ok. What's the difference between a Settlement with only one Building? I think those places would just be a small settlement

mokun:
> What's the difference between a Settlement with only one Building? I think those places would just be a small settlement

I suppose the life support requirements of a remote station can be tremendously curtailed, compared to a full-blown settlement, you know.

So we need only a minimally viable amount of air pressure and temperature to keep the equipment functioning.

But if a person is to enter into such a remote station, it would need to pump air into it to make it habitable or just good enough for a person donned with a EVA to repair or upgrade on the equipment.

As for robot, we could formulate a minimal set of requirement for a robot to perform basic maintenance tasks.

bevans2000:
Ok, I can see that but we just need a way to parameterise the life support setup. Creating a whole new concept is a lot of work.

PR #1596, opened 2025-06-28
Convert CookMealMeta to use Settlement backlog

### LINKED ISSUE #1571: Combine HotMeal & Dessert
## Situation
The simulation support the concept of Hotmeals and Dessert use 2 separate set of processing logic and associated classes. This is rooted down in 2 different Function being provided for 2 set of receipts that are actually fundamentally the same. This results in a lot of duplicate code.

In addition, both of the above are using Worker specific evaluations in MetaTask which has the potential for significant duplicate computation of scoring and selection.

## Proposal
The proposal here is to combine these into a single strategy for cooking dishes; whether they be Main dishes, Side Dishes or Dessert Dishes they all involved the same operations and actions. The receipt info is the same for all 3 categories so logically they can be combined.
It may be that whilst there is a single Function class to cook dishes, it make be parameterised with the category of dishes supported.
In addition the combined logic would also migrate over to the SettlementTask pattern allowing a Settlement wide demand to create dishes be calculated centrally.

There would also be consolidation in the UI layer.

## Approach
The approach is to merge Desserts with the Cooking logic set. Some class will be best renamed to better clarify their purpose, e.g. HotMeal becomes FoodDish that cover cover all 3 categories along with CookMeal becoming CookFood.
This is not a small task and there is a shortage of UnitTests in this area.

### LINKED ISSUE #1578: Convert CookMeal to Settlement Task pattern
There is a lot of duplicate logic applied whenever each Person/Robot is looking for new tasks. The majority of this logic is Settlement specific and hence is better matched to the ```SettlementTask``` approach.
The high level approach is:
1. Establish if the Settlement can cook ANY meal; this would prevent any backlog tasks
2. Create a ```RatingScore``` which reflects the cooking time schedule, i.e. meal times
3. For each Kitchen function create a SettlementTask that uses the common ```RatingScore``` and the "size" is the number of available chef slots at that Function

The Person assessment would look at the hunger rating. Whilst the Robot assessment would look to totally bias towards the Chef Bot; only that robot can do cooking.

bevans2000:
This PR converts the CookMealMeta to the Settlement backlog approach. This has performance advantages since many of the conditions to start cooking are central to the Settlement and hence the Backlog approach means the conditions are only evaluated. Secondary, it simplifies the code base and gives better visibility to the scoring of the Tasks.
The approach creates one TaskJob for each Cooking Building that can support more chefs.
The centralisation has allowed the scoring to be refined so a modifier is added based on the shortfall of meals in a Settlement.  

In addition, the Tasks associated with Farming & Malfunctions are moved into 'task' subpackage in a correct more specific parent. This is part of #1298.

Close #1578
Close #1571


mokun:
> [CodeFactor](https://www.codefactor.io/repository/github/mars-sim/mars-sim/pull/1596) — 4 issues fixed. 6 issues found.

Hope the issues reported by CodeFactor don't matter.

Issue #1550, opened 2025-03-23
Combine the Task package


bevans2000:
Move the sharable classes into the top-level Task package.
Keep Person specific Tasks under the person package

PR #1607, opened 2025-07-13
Fix numerous issues with Sites & Exploration

### LINKED ISSUE #1599: Improve the ability for Explored Locations to be used in Mining
The ability to run ```MiningMission``` is limited by the number of ```ExploredLocation``` in the correct state, e.g. has been claimed & explored as well as having concentrations. The explore Console Command can create new Explored Locations but just creates the vanilla state at a specific location that must have mineral concentration. The objective of this change is to make this easier to create locations that can be used by Mining.

The change proposed is:
- ```ExploredLocation.setClaimed``` will take a Settlement as argument; then isClaimed checks presence of  a Settlement. This means a Location must be claimed by a Settlement.
- Change the ExploreSite Task to claim the location for the Settlement of the Person
- Explore command becomes a Settlement console command
- Update explored add sub-command to the find suitable Location within a radius from the Settlement instead of a specific location. This can used the existing method in ```MineralMap``` to find a location with mineral concentration.
- Offer user what status; mark Explored & Claimed or Claimed (Not Explored) or Unclaimed
- If Claimed then it will be against the connected Settlement

bevans2000:
There are numerous changes but the main highlights are:

- MineralSite are now owned by a reporting authority instead of Settlement; allows Explored site to be shared for Mining
- Fix problem with ExploreSite EVA terminating and sending Workers into a zombie state
- Added a fallback to EVAOperatino to prevent Tasks being ended when a Worker is outside
- Correct problem with EquipmentInventory not actually checking for the correct EquipmentType
- Improve the information shown in Vehcile crew command

Ref #1550 

Close #1599


mokun (approved):
> MineralSite are now owned by a reporting authority instead of Settlement

Love that !

mokun:
Oh may be I did something to the ice and regolith probability value. Investigating them now.

Issue #1584, opened 2025-06-14
Create concept of MissionObjectives. Improve Mission Tool UI


bevans2000:
The various Mission types and subclasses each need a dedicated UI to show the purpose and outcome of a Mission. With the new generic MissionProject approach this is not possible as there are no specialised Mission classes. A generic means to represent the progress/objective of Mission is required.

The proposed solution is to create a new interface ```MissionObjective``` that represents the objective of a Mission, a Mission my have multiple. Each type of Mission can implement a dedicated class that holds the logic for that mission type. For example ```MissionCollectionObjective``` would be used for the collection mission and hold the counts of the various Resources collected. Essentially this is moving the logic into a dedicated class.

The benefit is in the UI a single Mission Window can be used that contains a flexible tab panel showing all MissionObjectives. Each objective subclass would have a dedicated panel but the selected panel would be date drive from the Mission instead of by the class type as currently, e.g. ```CollectionPanel```.

This approach would remove the need for different UI flow for the different Mission types; all Mission would be displayed according to their data not class. Thus this can also then support the generic MissionProject approach.

bevans2000:
The Mission Tool needs more reworking than expected to support this.

mokun:
@bevans2000 

Is this for the current implementation of Mission ? 

Should we do that after transitioning to the new Mission Project paradigm ?

mokun:
> The MissionObjective concept means the new style mission project can have multiple objectives if it has a mixture of steps.

So under the new paradigm of the mission project, we will enable having multiple objectives within a mission, right ?

> So eventually we can create a mission project that has a trading pages as well as an exploration phase on the way back, each will have a different mission objective

So is MissionObjective replacing the idea of a phase in Mission ?

bevans2000:
No Mission objective records the results of a Mission. Have a look on the branch which has an example for the Resource Collection missions. It will help a lot.
It might be better to call it MissionOutcome, haven't decided yet. Just applying the new concept to field study

mokun:
> No Mission objective records the results of a Mission... It might be better to call it MissionOutcome

Oh I see.

> each will have a different mission objective

So in the past, a mission has a fixed set of phases.

In the new mission project paradigm, a mission will have a flexible (customizable) set of phases or objectives, right ?

mokun:
> So the MissionObjective is where to record the results.

[...]

PR #1630, opened 2025-07-26
Convert Delivery mission to Objectives

bevans2000:
This PR converts the Delivery Mission to the new Mission Objectives pattern. In addition, it reuses the new standardised Negotiate Trade Task provided by the previous TradeMission conversion.

In addition, it reworks how Vehicles are selected for Missions by using the standard Comparator logic.

Ref: #1584


mars-sim-core/src/main/java/com/mars_sim/core/person/ai/mission/AbstractVehicleMission.java
```diff
@@ -232,31 +234,14 @@ protected void setInitialPhase(boolean needsReview) {
 	 */
 	private final boolean reserveVehicle() {
 		Collection<Vehicle> vList = getAvailableVehicles(getStartingSettlement());
-		List<Vehicle> bestVehicles = new ArrayList<>();
-
-		for (Vehicle v : vList) {
-			// CHeck if vehicle is usable
-			if (isUsableVehicle(v)) {
```
mokun, line 239: Good that you get rid of these layers of if else. They seemed to make a lot of duplicated effort in checking for a ready vehicle.

bevans2000: Yes I thought it would be better to tidy and simplify them. I like the change to use a Comparator<Vehicle> as well.

Issue #1614, opened 2025-07-19
Run processes on separate threads


mokun:
**Describe the bug**
- Currently, a lot of tasks are run on the master clock's thread.

**Expected Solutions**
- If an independent process needs to be run consistently as computing the outdoor temperature and solar irradiance, it should run onto a separate core for better thread management.
- Orbital movement, weather-related, lunar-related, and transport-related processes are suited to run on their own threads.

**Java classes involved**
- In `Simulation`, we have : 

```java
	/**
	 * Sends out the clock pulse instance.
	 *
	 * @param pulse the amount of clock pulse passing (in millisols)
	 */
	@Override
	public void clockPulse(ClockPulse pulse) {
		if (doneInitializing && !clockOnPause) {
			// Refresh all Data loggers; this can be refactored later to a Manager class
			DataLogger.changeTime(pulse.getMasterClock().getMarsTime());
			
			// Future: Will call each nation's timePassing(pulse) once per pulse
		
			lunarColonyManager.timePassing(pulse);
			
			orbitInfo.timePassing(pulse);
			
			weather.timePassing(pulse);

			surfaceFeatures.timePassing(pulse);

			unitManager.timePassing(pulse);

			transportManager.timePassing(pulse);
			
			// Pending save
			if (savePending != null) {
				saveSimulation(savePending, savePendingFile, saveCallback);
				saveCallback = null;
				savePending = null;
			}
		}
	}

```
So far, it seems only unitManager.timePassing runs on separate threads for each settlement as follows : 

```java

	/**
	 * Notifies all the units that time has passed. Times they are a changing.
	 *
	 * @param pulse the amount time passing (in millisols)
	 * @throws Exception if error during time passing.
	 */
	@Override
	public boolean timePassing(ClockPulse pulse) {
		if (pulse.isNewSol() || justLoaded) {
			if (factory != null) {
				// Compute reliability daily for each part
				factory.computePartReliability(pulse.getMarsTime().getMissionSol());
			}
			justLoaded = false;
		}

		if (pulse.getElapsed() > 0) {
			runExecutor(pulse);
		}
		else {
			logger.warning("Zero elapsed pulse #" + pulse.getId());
		}

		return true;
	}

```

**Specifications**
- pre-3.10.0 

**Additional context**
- The idea of this improvement came from discussing [this](https://github.com/mars-sim/mars-sim/issues/1610#issuecomment-3092151793)

bevans2000:
[...]

PR #1632, opened 2025-07-27
Provide different executor options for UnitManager

bevans2000:
This PR isolates the threading logic from UnitManager into a new set of classes that provide an option to change the executor behaviour. This is configured in a new property called 'executor-type' that can have 2 options:

1. 'shared' - This value uses a cached tread pool to executable the pulse in the Settlements. This is similar to the original approach but is allowed to grow and shrink the pool size as demand permits
2. 'thread' - This value uses a dedicated permanent Thread per Settlement. This removes the excessive logic of using Callables.

In addition, the dependencies on UnitManager and TransportManager have been reduced,
ref #1614


mokun:
@bevans2000 I wonder if there's a way to gauge the performance improvement of this virtual thread approach vs the old way.

Issue #1642, opened 2025-07-30
No construction of new buildings


bevans2000:
## Problem
There are no Construction Missions created automatically.  The Construction Mission meta has to find a site for working but creating a new Building is complex requiring a series of calculations that are failing or difficult to understand.
The placement of the new Building is complex and duplicates existing logic.
In terms of performance impact, each Person individually evaluates the chance of doing a ```ConstructionTask``` which involves scanning all Missions for a suitable candidate.
Lastly, the Construction Mission has a lot of logic that is specific to the Site and Stage. And in the reverse the Site duplicates data that is held in the Mission. The design is very tightly coupled resulting in missions being a Monster class.

## Solution
The solution is to simplify the creation of a new Building into a set of smaller phases that can be monitored and controlled by the user. Plus allowing more automatic creation as well as isolated testing.
The significant change is to make a Building queue in the ```ConstructionManager```.  This would be a list of items that represent where a new Building is to be built; it would be a combination of a ```LocalPosition``` & ```BuildingTemplate```. The latter being re-used from ```SettlementTemplate``` and ```ArrivalMission```. This represents the backlog of new Buildings to be created. The ConstructionManager will take responsibility of creating and selecting Sites. The mission will simply request the next best site to work on.
A new UI panel would be provided to manage and control the queue.
This would simplify the construction start because the user (or algorithm) would just select the type of Building and add it to the queue. The manager would then control creating the appropriate sequence of Construction Stages and base Site.

### Construction Mission Creation
This could be taken further by replacing the complex Construction Mission wizard and replacing it with a Building create dialog. The dailog would just require the user to select the Building type and a location. Then this would be added to the construction queue. The dialog would be started from the Settlement Map Tool after the user has already selected a position or direct from the Console.

In addition, the Construction Manager would be able to establish whether a Hallway is needed to connect the new Building to the existing Settlement.

The existing ConstructionMission wizard would then be a simple class of selecting Vehicles & People to start a Construction Mission. The site placement and selection is now the responsibility of the ConstructionManager. The site will not be directly created by the ConstructionMission; it will be created by the manager in response to a mission requesting an active site.

### Construction Site Creation
Based on the queue concept creating ConstructionSite takes on a different purpose. Currently a blank site is created and then the various properties are defined. This is open to mistake.
[...]

PR #1655, opened 2025-08-09
Redesign of Construction

bevans2000:
This is a PR that covers numerous changes all related to Construction. At a high level:

- Creation of a Construction Site is removed from the mission and now controlled by the ConstructionManager either directly by the user via a new Console command or via the building Queue logic.
- Construction Stage can only be created within a Site since a Site now has to have a final target Building at the outset. It can be possible to change this if the Stages permit.
- The Construction Mission wizard expects the user to only select an already created Construction Site.
- Placement of new Building is isolated into a separate class for reuse in the Supply mission (not done yet)
- Substantial increase to the Unit Tests
Ref #1642


mokun:
Glad that we finally have a construction wizard coming into shape !

mokun:
While we are writing codes that automate the construction steps and processes, I feel like we should consider player interactivity.

Can we begin building a standard set of dialog boxes that would pause the simulation and ask for players to check/uncheck options or a list of multiple choices of directions ? 

This way, players feel they are indeed in the loop for what happens in a settlement.

bevans2000:
Maybe later. We need to focus on completing what we start first to make sure it's working correctly but adding more stuff

mokun:
May as well merge it as soon as you're done. 

This way others can contribute to providing community feedback and future improvement.

mars-sim-core/src/main/java/com/mars_sim/core/building/construction/ConstructionManager.java
```diff
@@ -24,19 +23,68 @@
...
 
 /**
  * Manager for construction sites at a settlement.
  */
 public class ConstructionManager implements Serializable {
 
+	public static class BuildingSchedule implements Serializable {
+		// Private handler to schedule the activiation
+		private class Handler implements ScheduledEventHandler {
+			@Override
```
mokun, line 43: I'll learn to use `ScheduledEventHandler` elsewhere eventually.

bevans2000: This is a good example. It adds an event when the Building is scheduled to start work. When the execute method fires Inthe future, it clears the time delay

bevans2000:
> May as well merge it as soon as you're done.
> 
> Knowing that everything is work in progress, it's normal for anything missing.
> 
> I would expect others contribute to providing community feedback and future improvement right off the bat.

I might leave it because I'm away for nearly two weeks and something might break. But jump on the branch give it a try. The site console command can add to the queue. Not done the UI yet.
The mission meta scoring for construction needs reading some more because really only architect and engineers should start brand new sites. Currently, the first item is removed from the queue virtually immediately which seems odd but I haven't debugged it. Overall it looks pretty good and seems to be working well.

Issue #907, opened 2023-06-06
How to build communication, seismic, weather monitoring stations


mokun:
**Is your feature request related to a problem? Please describe.**
- What would it take to build communication, seismic, weather monitoring stations that belongs to a settlement ?

**Describe the rationale**
- As a settlement expands, it needs to track the changes or the impact made to its local environment. 
- Monitoring stations :
   - Enhance the understanding of the natural environment of Mars.
   - Collect on-site weather and seismic data, process them and relay them back to the settlement that owns it.
   - A mesh network of monitoring stations would provide valuable environment security to the health and well-being of a settlement.

- Communication Towers : 
   - Without communication towers, a settlement would have to rely on limited narrow bandwidth satellite comm for long-haul, such as connecting with remote rovers over hundreds of miles away or other settlements thousands of miles away.
   - Setting up remote communication towers along a mission corridor would pave the way for high bandwidth terrestrial 
communication with mission rovers.

**Describe the solution you desire to see**
- For players, use Mission Tool to create a *Building a monitoring station* mission
- For AI, use a weather debt score or a communication debt score to gauge the demand on when to dispatch such a mission.
- It is both a Building Mission and a Vehicle Mission. 
- It would require vehicles loaded with concretes or other building materials to be sent to a remote location
 
- For a small outside-only station, 
   - It only need to assemble large parts on-site and firmly secure the equipment to the ground and calibrate the instrument to function. 
   - It would need to place some kind of weather protective barrier around the equipment stations and instruments
   - It would need to periodically send settlers once in a while to check up what a robot cannot do or to perform upgrades
   - Passive remote stations would collect and relay data immediately back to the settlement.
   - Active stations would need to be furnished with computing equipment to process the data on site. 

- For a habitat-size station having indoor space, 
   - it would look like a small settlement and may require constructing a foundation, a frame and a building, much like a *Building Construction* mission.
   - it will need to have living space with life support built-in to house staff and researchers and it would need to house a robot to do on-site, regular maintenance

**Specifications  (please complete the following information):**
 -pre 3.6.0 build 8142

mokun:
@bevans2000 

I began a manual *Building Construction* mission. I picked 4 persons and a vehicle.

```java
01-Adir-01:386.550 (Info) AbstractMission : [Camp Kelper] Amy Kelly - Began organizing a Building Construction #5 mission.
[...]

PR #1663, opened 2025-08-21
Create DustStormForecaster.java

wonderinglostsoul44:
Dust storm forecaster

### DIFF
--- a/mars-sim-core/src/main/java/com/mars_sim/core/environment/DustStormForecaster.java
+++ b/mars-sim-core/src/main/java/com/mars_sim/core/environment/DustStormForecaster.java
@@ -1,0 +1,91 @@
+package org.mars_sim.msp.core.environment;
+
+import java.util.Arrays;
+
+/**
+ * Simple dust-storm forecaster using exponentially weighted moving average (EWMA)
+ * of atmospheric opacity (tau) and near-surface wind speed. Outputs a hazard index in [0, 1].
+ *
+ * This is deliberately simple and deterministic, meant as a building block for UI/EVA guards.
+ */
+public final class DustStormForecaster {
+
+    /** EWMA smoothing factor in (0,1]; larger = more reactive. */
+    private final double alphaTau;
+    private final double alphaWind;
+
+    /** Normalization constants (domain-expert tweakable). */
+    private final double tau90th;   // tau level that maps ~ to 0.9 hazard from tau alone
+    private final double wind90th;  // wind (m/s) mapping ~ to 0.9 hazard from wind alone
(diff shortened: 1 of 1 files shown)


bevans2000:
@wonderinglostsoul44 Hi, thanks for the contribution. How does the DustStormAdvisoryService get called by the existing Simulation engine?

mars-sim-core/src/main/java/com/mars_sim/core/environment/DustStormForecaster.java
```diff
@@ -0,0 +1,91 @@
+package org.mars_sim.msp.core.environment;
+
+import java.util.Arrays;
+
+/**
+ * Simple dust-storm forecaster using exponentially weighted moving average (EWMA)
```
mokun: @wonderinglostsoul44,

Appreciate your contribution and I like your proposal of applying exponentially weighted moving average (EWMA) for forecasting dust storm. 

For sure in near future, we could use some help from you in stringing together a good time series of weather data beyond the dust storm modeling.

mokun:
@bevans2000 

> How does the DustStormAdvisoryService get called by the existing Simulation engine?

While @wonderinglostsoul44 is figuring out the internal working of the simulation bit by bit, we could discuss some overarching ideas on having a local weather tower (a separate building function) for each settlement and how it would benefit the EVA ops, radiation exposure and temperature or dust storm monitoring, etc.

May be We can have `Command Center` (just being renamed recently, used to be `Command And Control`) providing the support with a certain radius coverage of these types of advisory service.

I imagine eventually the experiences of operating a weather tower should be translated into the knowledge repository of running construction missions of building [remote weather stations](https://github.com/mars-sim/mars-sim/issues/907) further out onto the surface of mars.

wonderinglostsoul44:
will do

wonderinglostsoul44:
The next steps would be to plug it in

Wire it into the sim loop. Wherever the environment/weather tick runs, call:

// one-time
DustStormForecaster forecaster = new DustStormForecaster();
DustStormAdvisoryService advisory = new DustStormAdvisoryService();
advisory.addListener(adv -> /* log or notify UI */ { /* ... */ });

// per-tick (after you have tau and wind)
forecaster.update(currentTau, currentWindMS, dtHours);
advisory.onTick(forecaster);


Decide thresholds and weights. If you have domain data, tune tau90th, wind90th, and tauWeight in the forecaster constructor to match Mars datasets you trust.

(Optional) Add tests. A small JUnit test ensuring monotonicity (hazard rises with tau/wind) and the severity banding would lock behavior in CI.

bevans2000:
We should also create a new map layer to show dust storms, both actual or forecaster.

PR #1666, opened 2025-08-22
Create SimClock.java

wonderinglostsoul44:
Deterministic by design: one hard tick; callbacks run in a stable order (tick, then registration sequence), so same seed → same outcome.

Minimal footprint: everything lives in one file, with nested types to avoid creating extra classes now.

Typed EventBus with priorities: subscribe(Class<T>, handler, priority) delivers synchronously, honoring isAssignableFrom for supertype listeners.

Future‑safe time types: SolInstant/SolDuration let you purge raw long/double time math gradually without adding more files.

Simple to wire: postLater and postAt use the same clock, so gameplay systems can schedule or react without spinning threads.

Next steps / options

Wire it in (MVP):

Create one SimClock at simulation bootstrap.

Replace ad‑hoc timers/loops with scheduleIn/At/Periodic.

Route system notifications through EventBus (e.g., ConstructionQueued, WeatherUpdated).

Auto‑tick event (optional): emit a TickEvent inside tickOnce() via bus.post(new TickEvent(nowTick())) once you’ve defined it.

Logging hook: swap System.err with your logger.

Gradual hardening: add unit tests for ordering guarantees; property‑tests for schedule math; then (if needed) split this file into SimClock.java + EventBus.java.

### DIFF
--- a/mars-sim-core/SimClock.java
+++ b/mars-sim-core/SimClock.java
@@ -1,0 +1,281 @@
+// package org.mars_sim.msp.core.sim; // ← adjust to your project package
+
+import java.time.Duration;
+import java.util.*;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.function.Consumer;
+
+/**
+ * Single-file simulation kernel:
+ * - Deterministic tick-based SimClock with a priority queue
+ * - Lightweight synchronous EventBus with priority + registration-ordered delivery
+ * - Minimal time types (SolInstant/SolDuration) to avoid bare longs
+ *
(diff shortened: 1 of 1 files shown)


wonderinglostsoul44:
if this one doesnt work reject, its more of an initial hard tick and deterministic scheduling skeleton draft that may or may not work at all

bevans2000:
I'm up for changing the clock logic because I think the current solution is way too complex. But it's not a small change as it integrates everywhere. Plus we have the concept MarsTime which it needs to support.
I would create a branch and try to apply this design change directly in the simulation.

PR #1667, opened 2025-08-22
Create ConstructionManager.java

wonderinglostsoul44:
Small, surgical changes: just a manager + events + test. No persistence, no UI, no heavy schemas yet.

Prevents a common class of bugs early: the tile reservation protects against two builds targeting the same spot, a source of hard‑to‑debug behavior later.

Deterministic & testable: IDs are monotonic, orders are FIFO, and events make it easy to hook UI/logic or write black‑box tests.

Extensible by design: the BlueprintValidator lets you drop in a real registry (JSON, XML, or code) without changing the manager API.

### DIFF
--- a/mars-sim-core/src/test/java/com/mars_sim/core/building/construction/ConstructionManager.java
+++ b/mars-sim-core/src/test/java/com/mars_sim/core/building/construction/ConstructionManager.java
@@ -1,0 +1,131 @@
+// package org.mars_sim.msp.core.construction;
+
+import java.util.*;
+import java.util.concurrent.atomic.AtomicLong;
+
+/**
+ * Minimal Construction MVP manager:
+ * - FIFO queue of construction orders
+ * - Tile reservation to prevent double-booking
+ * - Deterministic IDs and event emission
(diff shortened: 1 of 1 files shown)


mars-sim-core/src/test/java/com/mars_sim/core/building/construction/ConstructionManager.java
```diff
@@ -0,0 +1,131 @@
+// package org.mars_sim.msp.core.construction;
+
+import java.util.*;
+import java.util.concurrent.atomic.AtomicLong;
+
+/**
+ * Minimal Construction MVP manager:
+ * - FIFO queue of construction orders
```
mokun, line 8: I like making it FIFO for managing construction tasks and it makes sense.

bevans2000: This would clash with the existing ConstructionManager we have I think. It works very differently.
We have lots (an every growing) list of open issues that need help.

PR #1671, opened 2025-08-23
Update EVAMission.java

passingbyhelpingout1:
Ok lets see how this goes, feedback is welcomed, im hopeful this fixes this issue at least 

Why this / Key points

Prevents mission stalls: if a member “teleports” back to a settlement, they’re now removed from the EVA mission during phase end, letting isEveryoneInRover() become true and the mission progress.

Safe iteration: collects members to remove first, then calls removeMember(...) after the loop—no concurrent modification risks; reuses the mission’s own bookkeeping.

Next steps / options

If logs feel too noisy for rare edge cases, change logger.severe(...) to logger.warning(...) (throttle is already present).

Add a one‑line unit/integration check around EVA closeout to assert that a mislocated member no longer appears in getMembers() after checkTeleported().

### DIFF
--- a/mars-sim-core/src/main/java/com/mars_sim/core/person/ai/mission/util/VehicleStuckWatchdog.java
+++ b/mars-sim-core/src/main/java/com/mars_sim/core/person/ai/mission/util/VehicleStuckWatchdog.java
@@ -1,0 +1,197 @@
+/*
+ * Mars Simulation Project
+ * VehicleStuckWatchdog.java
+ * @date 2025-08-24
+ * Author: GPT-5 Pro (utility class)
+ *
+ * A tiny, self-contained watchdog to detect "no progress" while travelling and
+ * attempt staged recovery: (1) reroute, (2) skip micro-navpoint, (3) abort & RTB.
+ *
+ * Usage (example inside a travelling phase):
+ *
+ *   // As a mission field:
+ *   private final VehicleStuckWatchdog.State stuck = new VehicleStuckWatchdog.State("Regolith Prospecting");
+ *
+ *   // Inside performPhase() when TRAVELLING:
+ *   boolean recovered = VehicleStuckWatchdog.checkAndRecover(
+ *       stuck,
+ *       getVehicleCurrentCoordinates(),        // Coordinates of rover now
+ *       getCurrentMissionLocation(),           // Coordinates of target/navpoint
+ *       getPhaseDuration(),                    // phase-relative millisols "now"
+ *       () -> recalculateRouteToCurrentNavpoint(), // Runnable: rebuild/refresh route
+ *       () -> skipCurrentMicroNavpoint(),          // Runnable: advance to next navpoint
+ *       () -> abortAndReturnHome(),                // Runnable: end or RTB
+ *       getVehicle().getName()                     // tag for logging
+ *   );
+ *
+ * You only need to add this file. Call it where appropriate.
+ */
+
+package com.mars_sim.core.person.ai.mission.util;
+
+import java.io.Serializable;
+import java.util.concurrent.ThreadLocalRandom;
+
+import com.mars_sim.core.logging.SimLogger;
+import com.mars_sim.core.map.location.Coordinates;
+
+public final class VehicleStuckWatchdog {
+
+    // ---- Tuning knobs (units in comments) ----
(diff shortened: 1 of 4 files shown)


mokun:
> @passingbyhelpingout1 One thing you must do is make sure the Unit tests are passing with your change. We should not merge pull requests that have failing unit tests

@bevans2000 

It fails due to the old `ResourceUtilTest`.

> ResourceUtilTest.testIsInSitu:20 Ice is raw material

Would you like to do a pull on each of them to see if they work and if they improve anything ?

mars-sim-ui/src/main/java/com/mars_sim/ui/swing/unit_window/person/TabPanelHealth.java
```diff
@@ -994,48 +995,46 @@ else if (columnIndex == 2) {
...
+			final NavigableMap<Integer, Double> s = new TreeMap<>(circadian.getSleepHistory());
+			final NavigableMap<Integer, Double> e = new TreeMap<>(circadian.getExerciseHistory());
+
+			// 2) Build the union of sols actually present in either series
+			final TreeSet<Integer> allSols = new TreeSet<>(s.keySet());
+			allSols.addAll(e.keySet());
+			final List<Integer> newRows = new ArrayList<>(allSols); // earliest -> latest
+
+			// 3) Atomically swap the snapshots & row index
+			sleepTime = Collections.unmodifiableMap(s);
```
mokun, line 1032: We've been having problems with displaying sleep time, especially when running the simulation in high speed.

May be this will fix that problem of parsing through the sleep history.

@bevans2000 
Would you like to give that a try ?

mars-sim-core/src/main/java/com/mars_sim/core/person/ai/mission/util/VehicleStuckWatchdog.java
```diff
@@ -0,0 +1,197 @@
+/*
+ * Mars Simulation Project
+ * VehicleStuckWatchdog.java
+ * @date 2025-08-24
+ * Author: GPT-5 Pro (utility class)
```
mokun, line 5: This will be your first class file creation.

@bevans2000 

Do you like that ?

mokun:
For your [TabPanelHealth.java](https://github.com/mars-sim/mars-sim/pull/1671/commits/5ddc12392e34dc35b948e345a475d17766a6a818), I just merged the earlier updated version that you made.

In future, can you make a separate pull request (PR) when you aim at making an improvement on a different subject ?

mars-sim-core/src/main/java/com/mars_sim/core/person/ai/task/util/PersonTaskManager.java
```diff
@@ -54,6 +55,18 @@ public class PersonTaskManager extends TaskManager {
 
 	private transient Person person;
 
+	// ---------------------------------------------------------------------
+	// Cache rebuild throttling to reduce task churn and CPU spikes per agent
+	// ---------------------------------------------------------------------
+
+	/** Do not rebuild the task probability cache more often than this (millisols). */
+	private static final double MIN_REBUILD_INTERVAL_MSOLS = 1.0D;
```
mokun: Rebuilding the task cache does seem computationally intensive and excessive.

But rebuilding it on every 1 millisol may be too long.

A lot of tasks will get stale and outdated.

passingbyhelpingout1: with the way i have it, always use the latest ( if multiple the one at the bottom) as i may run 2-4 cycles of coding through and update, so essentialy the first one is a precursor.  But in the pull request, the actual changed code is always the latest version, so just merge the pull request

mokun:
The problem is that at this point, we would like to have the task cache rebuilt once a frame, not once every millisol.

That's why we want to hold it off for now.

But I do see in future, it's possible that at very high speed of the simulation, there's an advantage to rebuild all task caches for EVERY settler once per millisol, and not just for one or two settlers. This way, it would still ensure fairness for every person in the settlement.

@bevans2000, do you think having a timer can be useful for high speed situation ?

I'm not sure about introducing jitter. As good as it sounds, it also creates synchronization problems because conditions for each task can change quickly and in the next frame, the condition become stale and the cached meta task from the last frame is no longer updated.

PR #1673, opened 2025-08-23
More conversion to Mission Objectives and clean up

bevans2000:
This PR contains 2 areas of change:
- Emergency Supply Mission is convert to Mission Objectives.
- Rescue/Salvage vehicle mission converted

The RescueMission has had numerous corrections and fixes.

ref #1584

### DIFF
--- a/mars-sim-ui/src/main/java/com/mars_sim/ui/swing/tool/mission/GoodsTableModel.java
+++ b/mars-sim-ui/src/main/java/com/mars_sim/ui/swing/tool/mission/GoodsTableModel.java
@@ -7,7 +7,6 @@
 package com.mars_sim.ui.swing.tool.mission;
 
 import java.util.ArrayList;
-import java.util.Collections;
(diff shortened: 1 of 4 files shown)


mars-sim-core/src/main/java/com/mars_sim/core/mission/MissionVehicleProject.java
```diff
@@ -32,6 +34,7 @@
 public class MissionVehicleProject extends MissionProject
```
mokun, line 34: So when will we be truly using MissionVehicleProject ?

bevans2000: Hahah good question.
I realised that to follow a evolution not revolution approach the Mission Objectives conversation needs doing. This breaks the hard dependency between the Mission subclasses and the UI. 
I've just got the Salvage mission to conversation left.

PR #1674, opened 2025-08-24
Update TabPanelHealth.java

passingbyhelpingout1:
(Some Change Notes for this one)

Why this is worth doing

Correctness – hunger label showed thirst
The Hunger field accidentally displayed the thirst cache, which regularly misleads players about nutritional status and can influence task choices/AI triggers that users expect to observe from the UI.

Correctness – body mass deviation initialized from max energy bodyMassDev was initialized from getPersonalMaxEnergy() instead of getBodyMassDeviation(), causing mismatched values until the first refresh.

Consistency – performance metric
The “Performance” field was initialized from person.getPerformanceRating() but updated using condition.getPerformanceFactor(), producing inconsistent jumps. Now both use the same source (getPerformanceRating()).

Actionable Health Risks – sort by risk, cap safely The “Health Risks” area previously iterated an unsorted map and always took the first 10 enum entries. It now sorts by highest risk first, shows up to N rows safely, and won’t crash if there are fewer than 10 risks.

Sleep/Food history tables – safer zero‑data handling Minor guards ensure the tables compute row counts/offsets safely when data is temporarily empty (this ties to the recent “sleep time shows 0 msols” reports)

What this patch changes in practice

Players see the right numbers. Hunger now displays hunger; body mass deviation displays… body mass deviation.

No more inexplicable jumps in “Performance”. Both initial and refreshed values use person.getPerformanceRating(), so the number behaves predictably rather than flipping between two definitions.

Health risks are finally ranked. You see the most pressing risks first, which helps players (and QA) validate whether current routines (sleep, food, EVA exposure) are improving the right things.

Tables don’t “blink” to zero rows during transient data lulls. The guards keep row counts consistent while histories are warming up or temporarily empty (useful given the recent sleep‑table reports).  GitHub

How to test quickly

Open a person’s Health tab before/after the patch.

Confirm Hunger shows the same number you see in logs/tooltip for hunger (not thirst).

Confirm Performance stays consistent between initial paint and subsequent update() calls.

On a fresh game (or after clearing histories), open the Health tab immediately:

Sleep and Food tables should stay stable (0 or more rows, no exceptions).

Scroll the Health Risks list: the highest percentages should appear first; no errors if the list has <10 items.

### DIFF
--- a/mars-sim-ui/src/main/java/com/mars_sim/ui/swing/unit_window/person/TabPanelHealth.java
+++ b/mars-sim-ui/src/main/java/com/mars_sim/ui/swing/unit_window/person/TabPanelHealth.java
@@ -61,6 +61,9 @@ extends TabPanel {
 
 	private static final String HEALTH_ICON = "health"; //$NON-NLS-1$
 	
+	/** Max rows to render in the Health Risks list. */
+	private static final int RISK_ROWS = 10;
+
 	private static final String THIRTY_DAY = "30-Day";
 	private static final String ANNUAL = "Annual";
 	private static final String CAREER = "Career";
@@ -195,7 +198,7 @@ extends TabPanel {
 						
 		hungerCache = (int)condition.getHunger();
 		hungerLabel = conditionPanel.addTextField(Msg.getString("TabPanelHealth.hunger"), //$NON-NLS-1$
-										DECIMAL_MSOLS.format(thirstCache), null);
+										DECIMAL_MSOLS.format(hungerCache), null);
 		
 		maxDailyEnergy = (int)(condition.getPersonalMaxEnergy());
 		maxDailyEnergyLabel = conditionPanel.addRow(Msg.getString("TabPanelHealth.maxDailyEnergy"), //$NON-NLS-1$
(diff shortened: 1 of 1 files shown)


passingbyhelpingout1:
Oops this was meant to be part of patch 1, but considering some of the changes, maybe best to treat this as its own seperate submission.  If it fails CI tests i can easily tweak it

mars-sim-ui/src/main/java/com/mars_sim/ui/swing/unit_window/person/TabPanelHealth.java
```diff
@@ -225,7 +228,7 @@ protected void buildUI(JPanel content) {
 		muscleTorLabel = conditionPanel.addRow(Msg.getString("TabPanelHealth.muscle.tolerance"), //$NON-NLS-1$
 				DECIMAL_PLACES1.format(muscleTor));	
 		
-		bodyMassDev = Math.round(condition.getPersonalMaxEnergy()*10.0)/10.0;
+		bodyMassDev = Math.round(condition.getBodyMassDeviation()*10.0)/10.0;
```
mokun, line 231: a good catch !

mars-sim-ui/src/main/java/com/mars_sim/ui/swing/unit_window/person/TabPanelHealth.java
```diff
@@ -460,12 +463,15 @@ else if (colIndex < 4) {
...
 		List<HealthRiskType> riskList = new ArrayList<>(riskMap.keySet());
-		
-		Collections.sort(riskList);
-		
-		for (int i = 0; i < 10; i++) {
+		// Sort by descending risk value and cap the number of displayed rows.
+		riskList.sort((a, b) -> Double.compare(
+				riskMap.getOrDefault(b, 0D),
+				riskMap.getOrDefault(a, 0D)));
+		int rows = Math.min(RISK_ROWS, riskList.size());
```
mokun, line 470: Yes we could cap the risk and showcase those health risks that are very high that should warrant immediate medical attention.

@bevans2000 

We haven't tied these health risks to triggering any health problems just yet.

We may have to add new complaints based on these health risks. 

Do you want to do the honor of improving that in a new GitHub issue tracker ?

mokun:
@bevans2000 

Do you want to merge this pull ?

bevans2000 (approved):

bevans2000:
Just approved it.

mokun:
It failed a test. Rerunning it now to see if it goes away.

mokun (approved):

PR #1675, opened 2025-08-25
Update AbstractVehicleMission.java

passingbyhelpingout1:
**Ok lets see how this one goes, below are AI notes on the changes. **

Great—here’s another one‑file improvement that targets a real gameplay loop pain point we’ve seen reported in issues: rovers (or other ground vehicles) occasionally get “stuck” during the TRAVELLING phase and never make it home after an emergency, which then cascades into soft‑locks and mission bookkeeping weirdness.

Below is a drop‑in patch for com.mars_sim.core.person.ai.mission.AbstractVehicleMission that adds a tiny, low‑overhead stuck watchdog + recovery. It does three things:

Continuously samples vehicle progress (distance moved over a short look‑back window while the mission is in the TRAVELLING phase).

Detects “no real movement” (e.g., < 20–50 meters over ~0.05 sols) and tries up to 3 quick recoveries by rebuilding the travel phase/route—this is intentionally conservative and mission‑agnostic.

Fails safe if still stuck after the recoveries: ends the mission with a new status vehicleStuck, writes a clear mission log entry, and (optionally) requests an emergency beacon on the rover if such a method exists in your Rover/GroundVehicle class.

Why here?
The travel loop for rover missions routes through AbstractVehicleMission and subclasses (RoverMission, EVAMission, etc.). That makes AbstractVehicleMission a clean, central place to add a generic, mission‑type‑agnostic “is our vehicle actually making headway?” check without touching drivers, path planners, or the UI.

Why this helps the gameplay loop

Medical emergencies: if someone gets hurt mid‑EVA and the mission pivots to “return to base”, this guard ensures the rover actually moves. If the pathfinder/driver ends up oscillating or wedged, players won’t be left with a “mission that never ends.”

Low overhead: we track only a short deque of (time, position) samples and perform one distance calculation per tick—negligible cost versus navigation.

Fail‑safe termination: if all else fails, the mission won’t poison the UI or scheduler; it exits cleanly with an explicit vehicleStuck status the player can see in logs.

Tuning & shortcuts

You can tweak:

STUCK_LOOKBACK_MSOLS (window length): increase to be more conservative; decrease to react sooner.

STUCK_MIN_PROGRESS_KM (threshold): raise to be stricter; lower to avoid triggering on slow crawls.

STUCK_RECOVERY_MAX: how many rebuilds before failing safe.

How to test quickly

Start a rover mission and force a return (e.g., trigger a medical emergency or manually set a phase change).

Temporarily hack STUCK_MIN_PROGRESS_KM up to 1.0 and set STUCK_LOOKBACK_MSOLS to 5D to trigger recoveries quickly.

Observe logs:

“Vehicle appears stuck … Attempting route rebuild (1/3)”

If genuine no‑progress persists: “Ending mission as ‘vehicleStuck’.”

Restore thresholds to defaults after validating.

Relation to known issues

Community reports have shown missions halting after emergencies and vehicles failing to return while crew “teleport” or the mission remains in limbo. This patch adds a generic, non‑intrusive guard layer that helps those cases by guaranteeing forward progress or a clean end state.

### DIFF
--- a/mars-sim-core/src/main/java/com/mars_sim/core/person/ai/mission/AbstractVehicleMission.java
+++ b/mars-sim-core/src/main/java/com/mars_sim/core/person/ai/mission/AbstractVehicleMission.java
@@ -7,9 +7,11 @@
 package com.mars_sim.core.person.ai.mission;
 
 import java.text.MessageFormat;
+import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Comparator;
+import java.util.Deque;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


passingbyhelpingout1:
Ok use the most bottom version, as i did another pass even though the first one passed marvin CI Build with no issues that stopped merging, it did have quite a few yellow advisories.- I can run a few more passes comparing the code with the error logs and see if the AI cna further refine and remove as many of those yellow errors as it can.  And am in the process of doing that now, reducing the yellow error count with each pass - but i can do that without submitting any more committs just yet.  If this build goes through without any issues, then ill keep doing passes until it reaches a point where no more fixes cna be applied and then ill submit it as a new request.  Just want to see if this one is acceptible and can go through and then see if its enough or if i need to refine further

mars-sim-core/src/main/java/com/mars_sim/core/person/ai/mission/AbstractVehicleMission.java
```diff
@@ -622,15 +663,98 @@ else if (DISEMBARKING.equals(phase)) {
...
+		// Require a meaningful time window to avoid triggering on very fresh routes
+		double window = last.tMillisols - first.tMillisols;
+		if (window < (STUCK_LOOKBACK_MSOLS * 0.66)) return false;
+
+		double distKm = first.pos.getDistance(last.pos);
+		return distKm < STUCK_MIN_PROGRESS_KM;
+	}
+
+	/** Attempt to refresh driving state; after N tries, end mission as VEHICLE_STUCK. */
+	private void tryRecoverOrFail() {
```
mokun, line 782: @bevans2000 

This sounds like a good measure to alert the player/developer using log statements that the vehicle was stuck. 

It also turns on the emergency beacon after n numbers of attempts.

Since you're focusing on missions, would you like to give it a try ?

Issue #748, opened 2022-11-13
Problem with pulse


mokun:
**Describe the bug**
Calling the following L227 inside clearDown() in `MaintainGarageVehicle` should not result in NPE unless vehicle or garage is null : 
```
garage.removeVehicle(vehicle, false);
```

**To Reproduce (Please provide steps)**
1. Speed up to 4096x and wait to see

**Java classes involved**
MaintainGarageVehicle, MaintainVehicleMeta

**Stacktrace**
```
01-Adir-02:332.992 (Severe) UnitManager$SettlementTask : Problem with pulse on Kyocera: null
java.lang.NullPointerException
	at org.mars_sim.msp.core.person.ai.task.MaintainGarageVehicle.clearDown(MaintainGarageVehicle.java:227)
	at org.mars_sim.msp.core.person.ai.task.util.Task.endTask(Task.java:332)
	at org.mars_sim.msp.core.person.ai.task.util.Task.clearTask(Task.java:315)
	at org.mars_sim.msp.core.person.ai.task.MaintainGarageVehicle.<init>(MaintainGarageVehicle.java:69)
	at org.mars_sim.msp.core.person.ai.task.meta.MaintainVehicleMeta$GarageMaintenanceJob.createTask(MaintainVehicleMeta.java:45)
	at org.mars_sim.msp.core.person.ai.task.util.PersonTaskManager.createTask(PersonTaskManager.java:230)
	at org.mars_sim.msp.core.person.ai.task.util.TaskManager.startNewTask(TaskManager.java:511)
	at org.mars_sim.msp.core.person.ai.task.util.PersonTaskManager.startNewTask(PersonTaskManager.java:224)
	at org.mars_sim.msp.core.person.ai.Mind.lookForATask(Mind.java:310)
	at org.mars_sim.msp.core.person.ai.Mind.takeAction(Mind.java:253)
	at org.mars_sim.msp.core.person.ai.Mind.decideTask(Mind.java:186)
	at org.mars_sim.msp.core.person.ai.Mind.timePassing(Mind.java:123)
	at org.mars_sim.msp.core.person.Person.timePassing(Person.java:856)
	at org.mars_sim.msp.core.structure.Settlement.timePassing(Settlement.java:1104)
	at org.mars_sim.msp.core.structure.Settlement.otherTimePassings(Settlement.java:987)
	at org.mars_sim.msp.core.structure.Settlement.timePassing(Settlement.java:940)
[...]

Issue #158, opened 2018-11-28
java.lang.IllegalArgumentException: Task phase is null


Urwumpe:
**Describe the bug/issue**
An exception occurs repeatedly, also this exception is not written to log.

**Affected Area**
- EVAOperation.performMappedPhase
- LoadVehicleEVA.performMappedPhase

**Expected behaviors**
 - No exception
- All exceptions should be written to log

**Actual/Observed Behaviors**
```
java.lang.IllegalArgumentException: Task phase is null
        at org.mars_sim.msp.core.person.ai.task.EVAOperation.performMappedPhase(EVAOperation.java:192)
        at org.mars_sim.msp.core.person.ai.task.LoadVehicleEVA.performMappedPhase(LoadVehicleEVA.java:406)
        at org.mars_sim.msp.core.person.ai.task.Task.performTask(Task.java:528)
        at org.mars_sim.msp.core.person.ai.task.TaskManager.executeTask(TaskManager.java:378)
        at org.mars_sim.msp.core.person.ai.Mind.takeAction(Mind.java:228)
        at org.mars_sim.msp.core.person.ai.Mind.timePassing(Mind.java:178)
        at org.mars_sim.msp.core.person.Person.timePassing(Person.java:947)
        at org.mars_sim.msp.core.UnitManager.timePassing(UnitManager.java:2092)
        at org.mars_sim.msp.core.Simulation.clockPulse(Simulation.java:1530)
        at org.mars_sim.msp.core.time.MasterClock$ClockListenerTask.run(MasterClock.java:835)
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
        at java.base/java.lang.Thread.run(Thread.java:834)
java.lang.IllegalArgumentException: Task phase is null
        at org.mars_sim.msp.core.person.ai.task.EVAOperation.performMappedPhase(EVAOperation.java:192)
        at org.mars_sim.msp.core.person.ai.task.LoadVehicleEVA.performMappedPhase(LoadVehicleEVA.java:406)
        at org.mars_sim.msp.core.person.ai.task.Task.performTask(Task.java:528)
        at org.mars_sim.msp.core.person.ai.task.TaskManager.executeTask(TaskManager.java:378)
[...]

PR #1677, opened 2025-08-25
Update ResourceProcess.java

passingbyhelpingout1:
Ok AI reasons for changing the code, did a quick review myself, adjusted one or two things but it looks pretty solid, if it doesnt pass the CI tests , ill do a second pass to correct errors

What we’re fixing (in one file)

In many runs, resource processes (chemistry, recycling, food tech, etc.) keep waking up every tick even when they can’t run (no inputs, no power, or output storage is full). That hammers ResourceProcess.processResources() and inventory calls—both are among the top CPU hotspots—and produces annoying flip-flop behavior (“try → fail → try → fail…”) with no gameplay value.  GitHub
+1

The one-file patch

This patch adds an adaptive backoff & auto-suspend to ResourceProcess.java. When a process is blocked, it sleeps for a short time; repeated blocks escalate the cooldown (up to a cap). On success, the cooldown shrinks. Result: far fewer pointless ticks, smoother loops, less CPU. It also throttles log spam for repeated block reasons like “not enough storage” that frequently show up during heavy sims. (You can see these messages called out in issue logs.) Why this improves gameplay (and performance)

Fewer “thrash ticks.” When a process can’t run, it waits a short time before trying again—instead of burning CPU every single tick. This directly targets the known hotspot ResourceProcess.processResources() and inventory calls.  GitHub

Smoother loops. Factories stop flickering between on/off when storage is momentarily full or power dips. They “cool off,” then re-attempt—more human-like behavior that players read as intentional, not buggy.

Less log noise. Repeated “blocked” reasons are coalesced; you still get timely status, but without flooding. (Issue logs show these messages pile up during heavy activity.)  GitHub

Toggleable & safe. Set -Dmsim.process.backoff.enabled=false to revert to the old behavior instantly (handy for comparison tests).

Sanity checks & quick test plan (no other files required)

Smoke: Start a small settlement with a couple of continuous processes (e.g., soil generation, water recycling). Fill the relevant output storage to force OUTPUT_FULL. Observe that the process sleeps for ~3s, then 6s, 12s… (cap 20 min), and resumes as soon as space frees.

Load: In a large sim, watch CPU with VisualVM / JFR. You should see fewer samples in ResourceProcess.processResources() and inventory getters, aligning with prior observations.  GitHub

Logs: When a process is repeatedly blocked for the same reason, you’ll see one line every ~15s plus an occasional “suppressed N” roll-up, instead of dozens of lines per minute.

Regression: Pass -Dmsim.process.backoff.enabled=false and confirm behavior returns to current baseline.

Why this direction (and where it fits next)

The team’s been pushing toward simplifying/consolidating processing logic and improving manufacturing loops; this backoff mechanism complements that direction and is intentionally one-file only. It also aligns with the broader CPU-reduction effort discussed in recent issues.  GitHub
+2
GitHub
+2

If you want me to target a different single file next, good candidates (each with clear upside) are:

SimLogger.java – add message de-dup & rate-limit (global log readability boost, zero call-site changes).

EVAOperation.java – guard against null task phases + friendlier failure handling (stability; Issue #158).  GitHub

MaintainGarageVehicle.java – null-safety in clearDown() to stop rare NPEs (one-file fix; Issue #748).  GitHub

### DIFF
--- a/mars-sim-core/src/main/java/com/mars_sim/core/resourceprocess/ResourceProcess.java
+++ b/mars-sim-core/src/main/java/com/mars_sim/core/resourceprocess/ResourceProcess.java
@@ -30,7 +30,33 @@ public class ResourceProcess implements ScheduledEventHandler {
 	private static final double SMALL_AMOUNT = 0.000001;
 	// How often should the process be checked? 
 	private static final double PROCESS_CHECK_FREQUENCY = 5D; // 200 times per sol
-	
+
+	// ===== Adaptive Backoff (one-file gameplay upgrade) ======================
+	/** System property to disable backoff if needed (default: true). */
+	private static final boolean BACKOFF_ENABLED =
+			!"false".equalsIgnoreCase(System.getProperty("msim.process.backoff.enabled", "true"));
+
+	/** Minimum & maximum cooldowns between retries when blocked (in mSol). */
+	// With typical 1000 mSol ~= 1 sol, 5 mSol ~ 7.4 minutes sim-time; 15 mSol ~ ~22 minutes.
+	private static final double MIN_BACKOFF_MSOL = 5D;
+	private static final double MAX_BACKOFF_MSOL = 15D;
+
+	/** Log throttle window when repeatedly blocked (in mSol). */
+	private static final double LOG_THROTTLE_MSOL = 5D;
+
+	/** Backoff grows when we repeatedly hit the same block reason. */
+	private double currentBackoffMSol = MIN_BACKOFF_MSOL;
+	private double backoffRemainingMSol = 0D;
+	private double logThrottleRemainingMSol = 0D;
+	private int suppressedLogs = 0;
+
+	/** Why the process is blocked right now. */
+	private enum BlockReason {
+		NO_INPUT, OUTPUT_FULL, OTHER
+	}
+	private BlockReason lastBlockReason = null;
+	// ========================================================================
+
 	/**
 	 * Represents the internal state of the process.
 	 */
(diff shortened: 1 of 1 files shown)


mars-sim-core/src/main/java/com/mars_sim/core/resourceprocess/ResourceProcess.java
```diff
@@ -30,7 +30,33 @@ public class ResourceProcess implements ScheduledEventHandler {
 	private static final double SMALL_AMOUNT = 0.000001;
 	// How often should the process be checked? 
 	private static final double PROCESS_CHECK_FREQUENCY = 5D; // 200 times per sol
-	
+
+	// ===== Adaptive Backoff (one-file gameplay upgrade) ======================
+	/** System property to disable backoff if needed (default: true). */
+	private static final boolean BACKOFF_ENABLED =
+			!"false".equalsIgnoreCase(System.getProperty("msim.process.backoff.enabled", "true"));
```
bevans2000, line 37: We control feature toggles via the XML Configs. So add these setting to the Resources config class. This will allow it to be careful controlled.
Also it will give you some experience on how configuration is done

bevans2000 (changes requested):

mars-sim-core/src/main/java/com/mars_sim/core/resourceprocess/ResourceProcess.java
mokun: Not sure why @passingbyhelpingout1 suggests using System.getProperty() ?!

Before that, it'll need System.setProperty() and we don't use that.

passingbyhelpingout1:
Closing off, went on a tangent that isnt necessary anymore, some pull requests work some dont

passingbyhelpingout1:
Closing off, went on a tangent that isnt necessary anymore, some pull requests work some dont

Issue #1629, opened 2025-07-26
Create a LifeHistory class


bevans2000:
Currently a Person has a Job History and Role History plus the pending addition of recording Training.
It makes sense to combine these into a single record of life events called ```LifeHistory```. This will be a implementation using the existing ```History``` class to provide a timestamp record of ```LifeEvents```. At this stage the life event could just be captured as a String value.
Usign a dedicated new class, no matter how simple, provides good seperation.

bevans2000:
@passingbyhelpingout1 if are looking for something meaty to do have a look at this. It involves refactoring how we track job and role changes into a single History object. Needs a new UI tab panel to show life changes for a Person or Robot.
We can then expand it to cover other life events like training.

passingbyhelpingout1:
> [@passingbyhelpingout1](https://github.com/passingbyhelpingout1) if are looking for something meaty to do have a look at this. It involves refactoring how we track job and role changes into a single History object. Needs a new UI tab panel to show life changes for a Person or Robot. We can then expand it to cover other life events like training.

it might be a partial first pass only, was looking to tinker more around the sides to start with.  Even if the inital code gets rejected and  fine folks can take something from it to use later on im fine with that.

bevans2000:
Sorry, been busy.
I think it's more complicated that we need. The History<T> provides the changing of a value over time so this cna be used directly in the LifeHistory class. What we need it the LifeEvent class to be defined. 
Plus to complete the issue it need retrospectively applying to the existing code base. Ideal we need the UI panel as well.
Generally in our PR we try to deliver a completed change and not individual files. 
Plus we need UnitTests.

mokun:
@bevans2000 @Urwumpe 

It dawns on me that we should also have a ServiceHistory class for each Robot and Vehicle as well for tracking what have happened to them since they have been put into service.

Btw, I'd like to have the DeathInfo somehow being a part of this LifeHistory class.

And I imagine LifeHistory will chronicle all major events, including those prior to arriving on Mars. 

It's like the profile of a person that a commander will have access to. 

I even see that a person was selected because there is something in the past as shown in this LifeHistory class that caught the attention of the agency and therefore he/she was chosen to embark on this journey of being on Mars.

On the other hand, a tourist would have a very light profile.

bevans2000:
I consider this more like a Journal, everything interesting and Entity does. Not the low level Activity log used for Tasks.
It would make sense to have a single implementation for robot, person and vehicle

[...]

PR #1678, opened 2025-08-25
Create LifeHistory.java

passingbyhelpingout1:
Below is a drop‑in class that implements the “LifeHistory” concept proposed in Issue #1629—a single place to record time‑stamped “life events” (job changes, role changes, training, etc.) for a Person. It stores events locally and can optionally mirror each entry into your existing History implementation via reflection, so you don’t need to change imports or depend directly on History’s package/signature right away. This matches the intent in the issue: unify job/role/training into a single, time‑stamped history of life events (initially string‑based).

How this addresses the request

Single, unified log of “life events” for a Person—no more separate job/role/training lists. Matches the feature request to “combine these into a single record of life events called LifeHistory,” initially capturing the event “as a String value.”  GitHub

Time‑stamped using MarsTime, as the issue describes (use the project’s History concept for timestamps). This class records MarsTime on every entry and can mirror to your existing History using reflection, so you can wire it in immediately without chasing packages.  GitHub

Low‑risk, one-file drop‑in. No API breakage; you can keep legacy fields for backward compatibility and start double‑writing to LifeHistory before removing old structures.

https://github.com/mars-sim/mars-sim/issues/1629#issuecomment-3219228913 (first pass attempt to create a framework or a partial solution to this request)

### DIFF
--- a/mars-sim-core/src/main/java/com/mars_sim/core/person/history/LifeHistory.java
+++ b/mars-sim-core/src/main/java/com/mars_sim/core/person/history/LifeHistory.java
@@ -1,0 +1,240 @@
+/*
+ * Mars Simulation Project
+ * LifeHistory.java
+ * @date 2025-08-25
+ * Implements a unified, time-stamped history of "life events" for a Person
+ * (job changes, role changes, training, custom events), with optional bridging
+ * into the existing History class via reflection.
+ *
+ * Related issue: https://github.com/mars-sim/mars-sim/issues/1629
+ * The initial requirement is to capture events as strings with timestamps,
+ * consolidating separate job/role/training histories into one log. See issue #1629. 
+ *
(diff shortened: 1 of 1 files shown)


passingbyhelpingout1:
https://github.com/mars-sim/mars-sim/issues/1629#issuecomment-3219228913  (first pass attempt to create a framework or a partial solution to this request)

mokun:
@bevans2000 

I'll add this class file and play with it. 

This way, @passingbyhelpingout1 can get some ideas on how to incorporate that into our codebase.

bevans2000:
Ok but I asked for some changes. It's using  Reflection when I see no need. We should be reusing the existing History<T> to hold the entries. This issues is not really about how to stored the events, it's about what the event should be and adding the common UI and applying it to the various places in the code base.

PR #1679, opened 2025-08-25
Update ScientificStudy.java

passingbyhelpingout1:
Why not fix it in ScientificStudy instead?

That’s also a good fix (e.g., guard isCollaborativeResearchCompleted(Person) to return false if a person isn’t a collaborator instead of throwing). It was the root cause behind a known exception pattern in the past, and doing that inside ScientificStudy is worthwhile too. The advantage of this PersonTaskManager change is that it hardens the entire AI task loop against any meta that throws for any reason—present or future—without touching other subsystems. It’s the most leverage for a one‑file patch. (Background on the earlier study/collaborator exception path her

If you want a second, surgical alternative (still one file)

If you prefer to push the fix into the study code instead, you can harden just one method:

mars-sim-core/src/main/java/org/mars_sim/msp/core/science/ScientificStudy.java
Change: Early‑out false in isCollaborativeResearchCompleted(Person researcher) when the person isn’t in collaborators, avoiding IllegalArgumentException spam.

Guard ScientificStudy.isCollaborativeResearchCompleted(...)—also one file, fixes a frequent source of errors but narrower in scope.

Either way, you get a measurable reduction in exception churn and smoother ticks, with no behavior changes except that impossible/invalid tasks are quietly considered “not eligible right now.”

### DIFF
--- a/mars-sim-core/src/main/java/com/mars_sim/core/science/ScientificStudy.java
+++ b/mars-sim-core/src/main/java/com/mars_sim/core/science/ScientificStudy.java
@@ -40,13 +40,13 @@ public class ScientificStudy implements Entity, Temporal, Comparable<ScientificS
 	// POJO holding collaborators effort
 	private static final class CollaboratorStats implements Serializable {
 		private static final long serialVersionUID = 1L;
-		
+
 		double acheivementEarned = 0D;
 		double paperWorkTime = 0D;
 		double reseachWorkTime = 0D;
 		MarsTime lastContribution = null;
 		ScienceType contribution;
-		
+
 		CollaboratorStats(ScienceType contribution) {
 			this.contribution = contribution;
 		}
@@ -54,18 +54,18 @@ public class ScientificStudy implements Entity, Temporal, Comparable<ScientificS
 
 	/** default serial id. */
 	private static final long serialVersionUID = 1L;
-	
+
 	/** default logger. */
 	private static final SimLogger logger = SimLogger.getLogger(ScientificStudy.class.getName());
 
 	// Data members
 	/** The assigned study number. */
-	private int id;	
+	private int id;
 	/** Maximum number of collaborative researchers. */
 	private int maxCollaborators;
 	/** The difficulty level of this scientific study. */
 	private int difficultyLevel;
-	
+
 	/** The average amount of base work time (millisols) required for proposal phase. */
 	private double baseProposalTime;
 	/** The average amount of base work time (millisols) required for primary research. */
@@ -82,13 +82,13 @@ public class ScientificStudy implements Entity, Temporal, Comparable<ScientificS
 	private double primaryWorkDownTimeAllowed;
 	/** The average amount of downtime (millisols) allowed for collaborative work. */
 	private double collaborativeWorkDownTimeAllowed;
-	
+
 	/** The amount of proposal time done so far. */
 	private double proposalWorkTime;
 
 	/** The primary researcher */
 	private Person primaryResearcher;
-	
+
 	private StudyStatus phase;
 	private String name;
 	private ScienceType science;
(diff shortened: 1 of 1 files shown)


passingbyhelpingout1:
Ok the bottom (most recent) committ is the one to merge, the first one failed JAVA CI - Build testing but it has been fixed up to make sure it passes

mars-sim-core/src/main/java/com/mars_sim/core/science/ScientificStudy.java
```diff
@@ -608,16 +608,13 @@ public synchronized void addCollaborativeResearchWorkTime(Person researcher, dou
...
-	}
-
-	/**
-	 * Checks if all research in study has been completed.
-	 * 
-	 * @return true if research completed.
-	 */
+	public synchronized boolean isCollaborativeResearchCompleted(Person researcher) {
+   
+    if (researcher == null || !collaborators.containsKey(researcher.getIdentifier())) {
```
mokun: There's actually no need of doubling checking that.

If you go upstream, you'll find the 9 methods that call  isCollaborativeResearchCompleted() have the person as the collaborator already.

If you downstream, you'll also find the method getCollaboratorStats() that are also guarding against whether the person is in fact a collaborator by calling collaborators.get(researcher.getIdentifier()).

mars-sim-core/src/main/java/com/mars_sim/core/science/ScientificStudy.java
```diff
@@ -600,35 +598,37 @@ public synchronized void addCollaborativeResearchWorkTime(Person researcher, dou
...
+
 	private boolean isAllResearchCompleted() {
 		boolean result = true;
 		double targetTime = getTotalCollaborativeResearchWorkTimeRequired();
-		
+
 		for (CollaboratorStats c : collaborators.values()) {
-			if (c.reseachWorkTime >= targetTime)
+			// If ANY collaborator has NOT reached the target, overall research is not complete.
+			if (c.reseachWorkTime < targetTime)
```
mokun: @bevans2000 

It does sounds like if the research work time is smaller than the target time, then, this collaborative research is INCOMPLETE.

Shall we accept this PR ?

mars-sim-core/src/main/java/com/mars_sim/core/science/ScientificStudy.java
```diff
@@ -712,24 +712,24 @@ public synchronized void addCollaborativePaperWorkTime(Person researcher, double
...
+	 *
 	 * @return true if paper writing completed.
 	 */
 	private boolean isAllPaperWritingCompleted() {
 		boolean result = true;
 		double targetTime = getTotalCollaborativePaperWorkTimeRequired();
 		for (CollaboratorStats c : collaborators.values()) {
-			if (c.paperWorkTime >= targetTime)
+			// If ANY collaborator has NOT reached the target, overall paper writing is not complete.
+			if (c.paperWorkTime < targetTime)
```
mokun: @bevans2000 

Was it the same mistake we made here for writing paper ?

bevans2000: This does indeed look a mistake but hard to understand when this happened.
@passingbyhelpingout1 can you check the UnitTests and update them to catch the problem so it doesn't happen again please

mars-sim-core/src/main/java/com/mars_sim/core/science/ScientificStudy.java
bevans2000: Yes it looks good. I've asked if we can update the UnitTest
There was a lot of white space in this file 😨

bevans2000 (changes requested):
Can we see if the UnitTests need updating to trap the defect with the paper writing

bevans2000:
No problem. Don't create the pull request until you are ready for someone to look at it

passingbyhelpingout1:
> No problem. Don't create the pull request until you are ready for someone to look at it

All good, on a seperate note the orbital layer inital implementation code is now ready for review when u have time 😃👍

PR #1680, opened 2025-08-25
Add Construction Queue to the UI

### LINKED ISSUE #1642: No construction of new buildings

bevans2000:
This PR bring the Construction Queue feature to the UI in the Settlement Construction tab panel. It allows the user to add and remove Buildings from the construction Q.

In addition it improves the modelling of Construction Sites by:

- Holding a plan/phases of all the Construction Stages that should be performed on this site to create the target Building. This change simplifies the logic of the Construction and Salvage Missions. 
- The Site records with Mission actively working on site instead of a flag.

These 2 above change provide a better UI experience to the users.

@mokun Overall this now provides a much better experience for the user in terms of Construction.

close #1642

### DIFF
--- a/mars-sim-core/src/test/java/com/mars_sim/core/building/construction/MockMission.java
+++ b/mars-sim-core/src/test/java/com/mars_sim/core/building/construction/MockMission.java
@@ -22,169 +22,156 @@ class MockMission implements Mission {
 
     @Override
     public String getContext() {
-        // TODO Auto-generated method stub
         throw new UnsupportedOperationException("Unimplemented method 'getContext'");
     }
 
     @Override
     public void abortMission(String reason) {
-        // TODO Auto-generated method stub
         throw new UnsupportedOperationException("Unimplemented method 'abortMission'");
     }
 
     @Override
     public void abortPhase() {
-        // TODO Auto-generated method stub
         throw new UnsupportedOperationException("Unimplemented method 'abortPhase'");
     }
 
     @Override
     public boolean isDone() {
-        // TODO Auto-generated method stub
         throw new UnsupportedOperationException("Unimplemented method 'isDone'");
     }
 
     @Override
     public String getName() {
-        // TODO Auto-generated method stub
         throw new UnsupportedOperationException("Unimplemented method 'getName'");
     }
 
     @Override
     public void setName(String name) {
-        // TODO Auto-generated method stub
         throw new UnsupportedOperationException("Unimplemented method 'setName'");
     }
 
     @Override
     public Settlement getAssociatedSettlement() {
-        // TODO Auto-generated method stub
         throw new UnsupportedOperationException("Unimplemented method 'getAssociatedSettlement'");
     }
 
     @Override
     public String getFullMissionDesignation() {
-        // TODO Auto-generated method stub
         throw new UnsupportedOperationException("Unimplemented method 'getFullMissionDesignation'");
     }
 
     @Override
     public MissionLog getLog() {
-        // TODO Auto-generated method stub
         throw new UnsupportedOperationException("Unimplemented method 'getLog'");
     }
 
     @Override
     public Set<MissionStatus> getMissionStatus() {
-        // TODO Auto-generated method stub
         throw new UnsupportedOperationException("Unimplemented method 'getMissionStatus'");
     }
 
     @Override
     public MissionType getMissionType() {
-        // TODO Auto-generated method stub
         throw new UnsupportedOperationException("Unimplemented method 'getMissionType'");
     }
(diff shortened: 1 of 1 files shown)


mars-sim-core/src/test/java/com/mars_sim/core/building/construction/MockMission.java
```diff
@@ -0,0 +1,191 @@
...
+import com.mars_sim.core.person.ai.mission.MissionPlanning;
+import com.mars_sim.core.person.ai.mission.MissionStatus;
+import com.mars_sim.core.person.ai.mission.MissionType;
+import com.mars_sim.core.person.ai.task.util.Worker;
+import com.mars_sim.core.project.Stage;
+import com.mars_sim.core.structure.ObjectiveType;
+import com.mars_sim.core.structure.Settlement;
+import com.mars_sim.core.time.MarsTime;
+
+class MockMission implements Mission {
```
mokun, line 21: Great that you're adding MockMission for testing missions

mars-sim-core/src/main/java/com/mars_sim/core/building/construction/ConstructionManager.java
```diff
@@ -319,6 +297,15 @@ public void addBuildingToQueue(String buildingType, MarsTime when) {
 		}
 	}
 
+	/**
+	 * Remove a previously added building from the queue
+	 * @param item Schedule to be removed
+	 * @return Was teh item found and removed
+	 */
+	public boolean removeBuildingFromQueue(BuildingSchedule item) {
+		return queue.remove(item);
```
mokun, line 306: So, will the task from this queue be FIFO ?

Issue #1665, opened 2025-08-22
Memory leak when using zoom slider in Settlement Map


mokun:
**Describe the bug**
- Whenever I roll the mouse roller or drag the zoom slider in the Settlement Map to change the magnification of the buildings, the memory meter will rise in the amount of the heap memory. The maximum will not come down and be clear. 
- Closing the settlement map doesn't not vacate the memory being used.

**Discuss any potential problems you have and have not ruled out**
- We do not remove any mouse listeners in `SettlementTransparentPanel`. It may or may be contribute to problem.
- This happens to a small 4-person settlement.

**Java classes involved**
- `SettlementTransparentPanel`, `SettlementMapPanel`
 
**To Reproduce (Please provide steps)**
- Steps to reproduce the behavior:
1. Open Settlement Map
2. Drag the scale slider up and down 
3. Watch the max memory to go up from 200 MB to 2000 MB within a few seconds in the Memory meter on the right bottom of the Main Window.
4. Even after closing the Settlement Map, the max memory won't go down.

**Screenshots**
- Initially it was only 94 of 166 MB 

<img width="791" height="812" alt="Image" src="https://github.com/user-attachments/assets/1359b886-c88c-42a8-886e-ac1cb066ea47" />

- As we slide the scale up, it increases to ~200 of 300 MB

<img width="788" height="811" alt="Image" src="https://github.com/user-attachments/assets/40bceb06-44d6-43ac-9a6d-1d8c1c5ab76c" />

- The max memory would continue to go up as we use the slider up and down. Soon, it reaches 2000 MB.

<img width="788" height="813" alt="Image" src="https://github.com/user-attachments/assets/f3a0c015-48d2-43b3-8ae3-abe387c49203" />

**Specifications**
- pre-3.10.0 

**Related Issues and Discussions**
- It may be related to the recent increase of usage of high fidelity svg images in Lander Hab and Compact Quarters. See https://github.com/mars-sim/mars-sim/issues/1187

bevans2000:
The memory not going when the window is closed is probably due to Unit listeners not being released or the fact the tool windows never get destroyed. We just close the window, not release it.
We release Unit windows but not tools.

wonderinglostsoul44:
Stop the Settlement Map zoom‑slider memory creep

Why: The zoom slider fires a torrent of change events and the map panel appears to keep listeners around after the window closes—classic Swing leak patterns. 
[...]

PR #1681, opened 2025-08-25
Update Mind.java

passingbyhelpingout1:
What this does, in gameplay terms

Minimum commitment window (default 15 millisols ≈ 22 minutes): once a colonist starts a task (e.g., Sleep), they stick with it for that window if the task remains valid.

Debounce: don’t rebuild probabilities / re‑pick tasks faster than a tiny jittered cooldown (cooperates with your existing PersonTaskManager throttle).

Hard overrides still win: emergencies, EVA/mission directives, or a task reporting “cannot continue” will immediately break the commitment.

Result: Sleep sessions look like real naps/sleeps instead of flickering; Eat/Drink flapping drops; CPU spikes from mass retasking diminish.

Why this is safe

No new dependencies: only uses MarsTime, TaskManager, and Task methods you already have (getTask(), startNewTask(), canContinue(), endTask()).

Doesn’t block critical flows: shouldForceRetask() keeps the same behavior when a task can’t continue, a mission forces a switch, or an emergency occurs.

Tunable: operators can tweak -Dmars.sim.task.session.min=… at runtime to calibrate stickiness per save.

Expected impact

Fixes the “0 ms sleep” flicker by preventing immediate post‑start re‑rolls unless Sleep reports it can’t continue. This addresses #1665 symptoms at the control point where churn happens.  GitHub

Calms global retask storms during peak sim speeds (fewer simultaneous re‑picks).

Improves believability: humans don’t reconsider every minute once they’ve laid down to sleep.

How to verify (quick)

Start a scenario with many colonists; let the clock run fast.

Compare task switches per sol before/after; Sleep tasks should last at least ~15 msols in the absence of emergencies.

Observe logs: you’ll see occasional “Forced retask -> …” when an override occurs, and fewer “picked new task” bursts right after a task starts.

Why not change Sleep itself?

We can also add hysteresis and a minimum sleep block inside Sleep.java, but the biggest churn win comes from centralizing the rule in Mind so it automatically benefits every task (Sleep, Eat/Drink, Hygiene, etc.).

### DIFF
--- a/mars-sim-core/src/main/java/com/mars_sim/core/person/ai/Mind.java
+++ b/mars-sim-core/src/main/java/com/mars_sim/core/person/ai/Mind.java
@@ -1,12 +1,13 @@
 /*
  * Mars Simulation Project
  * Mind.java
- * @date 2022-08-06
+ * @date 2025-08-25 (patched)
  * @author Scott Davis
  */
 package com.mars_sim.core.person.ai;
 
 import java.io.Serializable;
+import java.util.concurrent.ThreadLocalRandom;
 
 import com.mars_sim.core.UnitEventType;
 import com.mars_sim.core.logging.SimLogger;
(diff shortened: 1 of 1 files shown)


passingbyhelpingout1:
Ok its still generating some issues, , ill revisit this one later today when i have time or tomorrow, so dont push through unless someone can fix the errors, ill get back to them, i did attempt to fix two other pull requests so they might be worth revisiting at some point, no rush

PR #1684, opened 2025-08-26
Update PersonTaskManager.java

passingbyhelpingout1:
What drives the day‑to‑day gameplay here

Tasks & Meta‑Tasks: Every settler picks from 50+ Task types; selection is driven by 40+ “Meta” classes that compute the probability of doing a task at that moment (e.g., SleepMeta → Sleep). This is the heart of the gameplay loop.  GitHub

Work shifts: Shifts determine what tasks are “on offer” (work vs non‑work vs any hour) and how people rotate. This strongly shapes colony rhythms.  GitHub

Clock & unit manager: The MasterClock ticks the sim; Simulation.clockPulse() fans out to managers; UnitManager dispatches per‑unit work. Stack traces show this call chain and where task selection happens.  GitHub

Two pain points show up in issues/discussion history:

Meta‑task blowups during probability calculation (e.g., when a person isn’t actually a collaborator yet) can throw and bubble through the tick loop. This causes noisy logs, wasted CPU, and “task thrashing” that harms pacing. (See the long stack trace with PersonTaskManager.rebuildTaskCache and PerformLaboratoryExperimentMeta.getProbability.)  GitHub

Mission emergency behavior (teleport/stuck edge cases) points to robustness gaps under stress, but that often spans multiple files.  GitHub

Given your “one file only” constraint, the safest, high‑impact target is the task selection hot path.

2) One‑file patch: “Task Cooldown & Guard Rails” in PersonTaskManager.java

What it does (in plain English)

Adds a cooldown for meta‑tasks that just failed during getProbability(...) (e.g., the “not a collaborator” case). Instead of exploding every tick (and being considered again a millisecond later), they’re temporarily dampened so settlers try alternatives first.

Wraps meta‑task probability evaluation with fail‑open guard rails (catching IllegalArgumentException/IllegalStateException/RuntimeException) so one bad meta can’t tank the whole tick.

Keeps the failed meta in the cache (with sharply reduced weight) so it can naturally come back when conditions change, avoiding starvation.

All of this happens inside one file: mars-sim-core/src/main/java/org/mars_sim/msp/core/person/ai/task/utils/PersonTaskManager.java (class and package confirmed by issue stack traces).  GitHub

Why this helps gameplay

Settlers stop thrashing on impossible or not‑yet‑valid tasks; they move on to the next best thing.

Fewer red herring exceptions in logs; lower CPU churn; smoother colony cadence.

Directly addresses the repeated “not a collaborative researcher” failure pattern seen in the wild.

Testing it quickly

Run a scenario where science tasks are available. The repeated “not a collaborator” path in the issues is easiest to reproduce (Areology field or lab work). When that meta throws, you should now see one FINE log and then silence for ~200 pulses while colonists pick other tasks. (The offending meta re‑enters naturally later.)  GitHub

Watch the Mission/Task monitor: you should see fewer rapid fire task flips and no exception cascade in the console.

Tune FAILURE_COOLDOWN_PULSES to taste (e.g., 50–400) to balance responsiveness vs. stability.

Why only one file?
Everything happens in the task selection loop (PersonTaskManager). No other classes or signatures need to change.

### DIFF
--- a/mars-sim-core/src/main/java/com/mars_sim/core/person/ai/task/util/PersonTaskManager.java
+++ b/mars-sim-core/src/main/java/com/mars_sim/core/person/ai/task/util/PersonTaskManager.java
@@ -7,6 +7,8 @@
 package com.mars_sim.core.person.ai.task.util;
 
 import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 
 import com.mars_sim.core.data.RatingScore;
 import com.mars_sim.core.logging.SimLogger;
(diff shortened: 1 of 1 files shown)


passingbyhelpingout1:
Resubmitted a slightly different approach that is a little bit less CPU intensive so ill close this one off, the new one is the latest approach with a few tweaks and different approach

PR #1686, opened 2025-08-27

passingbyhelpingout1:
I tracked the error you pointed at in that PR thread down to a stray call in SettlementTransparentPanel.removeNotify():

The code calls removeMouseWheelListener(mwl); but there is no variable named mwl in scope. That produces a compile error (“cannot find symbol: variable mwl/interface expected here” depending on where the compiler syncs).

The same block already removes the listener correctly from the mapPanel using the field mouseWheelListener, so the mwl call is not only wrong but redundant. You can see the problematic line in the review snippet for that PR (the “Files changed” view shows the removeMouseWheelListener(mwl); line, followed by correct removal from mapPanel).  GitHub
+1

Below is a surgical patch that removes the bad call and makes the listener cleanup consistent. I also null‑out the field after removal to avoid stale references.

Why this fixes it:

removeMouseWheelListener(mwl); referenced a non‑existent local variable. The actual listener instance is stored in the mouseWheelListener field; removing via that field on the same component it was added to (mapPanel) is correct and sufficient. The redundant and invalid mwl call is deleted, compile error goes away, and cleanup is consistent. The PR’s code review snippet shows this exact region with the stray call.

Notes

No behavior change beyond removing the invalid line; the cleanup still occurs where it should (on mapPanel), which is what the PR intended. The problematic line is visible in the PR’s diff snippet.  GitHub

If you still see a compile error, it likely means there’s another stale reference to mwl elsewhere in the file. Search for mwl and delete/replace as above.

### DIFF
--- a/mars-sim-ui/src/main/java/com/mars_sim/ui/swing/tool/settlement/SettlementTransparentPanel.java
+++ b/mars-sim-ui/src/main/java/com/mars_sim/ui/swing/tool/settlement/SettlementTransparentPanel.java
@@ -793,16 +793,21 @@ public class SettlementTransparentPanel extends JComponent {
 
     @Override
     public void removeNotify() {
-        super.removeNotify();
-        if (zoomSlider != null && zoomListener != null) {
-            zoomSlider.removeChangeListener(zoomListener);
+        try {
+            if (zoomSlider != null && zoomListener != null) {
+                zoomSlider.removeChangeListener(zoomListener);
+            }
+            if (zoomDebounce != null) {
+                zoomDebounce.stop();
+            }
+            // Remove from the same component we added to (mapPanel), not 'this'
+            if (mouseWheelListener != null && mapPanel != null) {
+                mapPanel.removeMouseWheelListener(mouseWheelListener);
+                mouseWheelListener = null; // avoid stale reference
+            }
         }
-        if (zoomDebounce != null) {
-            zoomDebounce.stop();
-        }
-        // Remove from the same component we added to (mapPanel), not 'this'
-        if (mouseWheelListener != null && mapPanel != null) {
-            mapPanel.removeMouseWheelListener(mouseWheelListener);
+        finally {
+            super.removeNotify();
         }
     }
 


passingbyhelpingout1:
As requested here is another pass on this one, let me know if it needs further work..  😃

mars-sim-ui/src/main/java/com/mars_sim/ui/swing/tool/settlement/SettlementTransparentPanel.java
```diff
@@ -793,16 +793,21 @@ private void convertTo(int value) {
 
     @Override
     public void removeNotify() {
```
mokun, line 795: @passingbyhelpingout1 

This is a small thing. I could still accept the PR.

You recommended adding this method in the past. But it actually didn't do anything in our use case.

I just tested it. This method never gets called when SettlementTransparentPanel (that extends JComponent) was destroyed.

Perhaps it's because SettlementTransparentPanel always goes hand in hand with SettlementMapPanel. When SettlementMapPanel is destroyed, it won't have the time to call removeNotify.

passingbyhelpingout1:
ok just did a quick patch, check the file again and see if that resolves those 10 or 15 lines that were causing the conflict

mars-sim-ui/src/main/java/com/mars_sim/ui/swing/tool/settlement/SettlementTransparentPanel.java
```diff
@@ -1290,3 +1290,4 @@ public void destroy() {
 
     }
 }
+
```
mokun, line 1293: what's changed ?

passingbyhelpingout1: i think it was just a change to make it CI test compatible.  The conflict was with the CI verification, nothing that should efffect gameplay, should be safe to merge

PR #1688, opened 2025-08-28

passingbyhelpingout1:
What I fixed

Root cause

In SettlementTransparentPanel.removeNotify(), the code invoked:

removeMouseWheelListener(mwl);

but there is no variable named mwl declared in that scope, which yields a “cannot find symbol” compile error. Meanwhile, the class already has a field named mouseWheelListener and correctly removes it from mapPanel. The mwl call is both invalid and redundant.  GitHub

Fix

Delete the invalid call to removeMouseWheelListener(mwl);.

Keep the real cleanup: if both mapPanel and mouseWheelListener are non‑null, remove the listener from mapPanel.

Null out the mouseWheelListener field after removal to avoid stale references.

Call super.removeNotify() in a finally block to ensure it runs exactly once even if cleanup throws, which also addresses the lifecycle consistency concern noted in that same PR discussion.  GitHub

This change is intentionally surgical and limited to one file.

hy this is correct

Fixes the compile error: Eliminating the reference to the non‑existent mwl resolves the “cannot find symbol” error highlighted in the PR discussion. The removal logic now refers only to mouseWheelListener, which is the actual field.  GitHub

Consistent lifecycle: Wrapping listener cleanup in try and calling super.removeNotify() in finally ensures superclass logic is executed exactly once and not skipped if an exception occurs — matching the recommended pattern mentioned in the same PR thread.  GitHub

Avoids stale references: Nulling out the field after removal avoids dangling references that could impede GC or cause re‑entrancy surprises later.

### DIFF
--- a/mars-sim-ui/src/main/java/com/mars_sim/ui/swing/tool/settlement/SettlementTransparentPanel.java
+++ b/mars-sim-ui/src/main/java/com/mars_sim/ui/swing/tool/settlement/SettlementTransparentPanel.java
@@ -1333,9 +1333,12 @@ public class SettlementTransparentPanel extends JComponent {
         }
         zoomListener = null;
 
-        // Remove from the same component we added to (mapPanel), not 'this'
-        if (mouseWheelListener != null && mapPanel != null) {
-            mapPanel.removeMouseWheelListener(mouseWheelListener);
+        // Defensive unregistration from both this panel and the mapPanel (safe no-ops if not present)
+        if (mouseWheelListener != null) {
+            removeMouseWheelListener(mouseWheelListener);
+            if (mapPanel != null) {
+                mapPanel.removeMouseWheelListener(mouseWheelListener);
+            }
         }
         mouseWheelListener = null;
 


passingbyhelpingout1:
im going through the marvin errors that pop up during the checks that run on each file as part of the CI process, ive noticed well over 1000 error messages.  So switching tack and will attempt to work through and fix those small bugs from i guess years of code buildup, they generate warnings and arent usually enough to stop the checks going through - so there might be some unusual committs made in terms of very small changes but i will attempt to reduce some of that small error debt that builds up and might reduce that yellow error total going forward

mokun:
After many change, the class file is becoming messy. 

I've just selectively incorporated some of changes you recommended on a commit to see if it removes some memory leak. 
Thank you. 

I'll close this PR now.

passingbyhelpingout1:
> After many change, the class file is becoming messy.
> 
> I've just selectively incorporated some of changes you recommended on a commit to see if it removes some memory leak. Thank you.
> 
> I'll close this PR now.

all good even if you can take some tidbits and snippets out of what ive put up, its a starting point, ill delete this branch

PR #1692, opened 2025-08-28
Update PhysicalCondition.java

passingbyhelpingout1:
Gameplay Loop: Automatic Simulation Halt on Colony Extinction (PhysicalCondition.java)

Issue: The simulation currently continues running even if all settlers die. This leaves the game in a stagnant state without a clear end condition for the player’s colony. In terms of gameplay loop, there should be a fail-state that is recognized by the game (game over when the colony is gone).

Improvement: Implement an automatic “game over” detection – when the last human colonist dies, the simulation will pause and log a message. This gives the player feedback that the colony has perished, effectively ending the gameplay loop in a clear way.

We implement this in the PhysicalCondition.recordDead(...) method, which is called when a person dies, by checking if any colonists remain alive. If none remain, we pause the master clock (freezing the simulation) and log a severe message.

Explanation: When a person dies, we gather their settlement and use UnitManager to get all people in the simulation (via getPeople() GitHub
). For each living person in the same settlement, we call addStress(15.0) on their PhysicalCondition to reflect the emotional trauma of losing a crewmate. The addStress method will increase their stress (capped at 100) and trigger a STRESS_EVENT update GitHub
, potentially leading to panic attacks if stress becomes extreme. This deepens the simulation by modeling morale and psychological effects of fatalities.

Next, we count remaining alive people. If none (aliveCount == 0), we invoke Simulation.instance().getMasterClock().setPaused(true, true) to pause the game, and log a severe message indicating the colony’s demise. Pausing the MasterClock effectively halts all time progression (similar to the user pausing the game), which serves as a game-over state. The log (and any on-screen console/UI bound to it) will notify the player that all settlers are dead and the simulation is stopped. This provides a clear end to the gameplay loop when the colony fails, rather than letting the simulation run indefinitely with no settlers.

(Note: Simulation.instance() provides access to the singleton Simulation, and from it we obtain the UnitManager and MasterClock. The above code assumes these static accessors exist as in the codebase (they are used elsewhere, e.g. Role.java uses Simulation.instance().getMasterClock()

### DIFF
--- a/mars-sim-core/src/main/java/com/mars_sim/core/person/PhysicalCondition.java
+++ b/mars-sim-core/src/main/java/com/mars_sim/core/person/PhysicalCondition.java
@@ -15,6 +15,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.logging.Level;
 
 import com.mars_sim.core.LifeSupportInterface;
@@ -33,6 +35,7 @@ import com.mars_sim.core.person.ai.task.util.Task;
 import com.mars_sim.core.person.ai.task.util.ExperienceImpact.PhysicalEffort;
 import com.mars_sim.core.person.health.Complaint;
 import com.mars_sim.core.person.health.ComplaintType;
+import com.mars_sim.core.person.health.CuredProblem;
 import com.mars_sim.core.person.health.DeathInfo;
 import com.mars_sim.core.person.health.HealthProblem;
 import com.mars_sim.core.person.health.HealthProblemState;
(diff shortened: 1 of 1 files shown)


mars-sim-core/src/main/java/com/mars_sim/core/person/PhysicalCondition.java
```diff
@@ -1550,6 +1563,20 @@ public void recordDead(HealthProblem problem, boolean triggeredByPlayer, String
 		// Add the person's death info to the postmortem exam waiting list
 		// Note: what if a person died in a settlement outside of home town ?
 		medicalManager.addPostmortemExam(person.getAssociatedSettlement(), deathDetails);
+
+		// ===== Colony-extinction detection and automatic pause =====
+		if (wasAlive) {
+			int remaining = ALIVE_COLONIST_COUNT.decrementAndGet();
+			if (remaining <= 0 && GAME_OVER_TRIGGERED.compareAndSet(false, true)) {
+				try {
+					master.setPaused(true, true);
```
bevans2000, line 1572: We shouldn't be stopping the clock from inside a logic class like this, only from controllers.

bevans2000:
We can't take this because it's not solving a problem. No people is not an end case, it is an unlikely but is a valid case

PR #1695, opened 2025-08-29
Create ResourceProcessPlanner.java

passingbyhelpingout1:
What it adds: a VP‑aware, power‑aware, hysteresis‑stabilized auto‑controller that starts/stops resource processes to meet stock targets (e.g., water, oxygen, metals, glass), using the sim’s Value Point (VP) economics so choices align with settlement needs.  GitHub

Why it matters: it turns raw gathering → refining → useful outputs into a smooth loop that avoids flip‑flopping and wastes less power/crew. It also plays nicely with the mineral/regolith → ore flows your wiki describes.  GitHub

One file only: everything (builder, scoring, hysteresis, state) lives in one Java file. You wire it up with a few lambdas; no imports from mars‑sim classes are required (works with both legacy org.mars_sim.msp.core… and newer com.mars_sim.core… packages because we don’t hard‑code types).

Data‑driven friendly: it respects your XML‑driven configs—targets and process I/O/rates are passed in from your existing loaders/utilities.

What this single file gives you

Automated resource processes that follow VP & needs The controller reads per‑resource Value Points (VP) and steers production toward what the settlement values most (water, oxygen, food, metals, glass, etc.), which is exactly how your economics model is intended to drive behavior.  GitHub

Stable behavior (no flapping)

Hysteresis bands around each target (e.g., 15%) prevent spammy start/stop when stocks hover near thresholds.

Min run ticks keep a process on long enough to be worthwhile before reconsidering.

Smarter trade‑offs

Power‑aware: Processes with high kW draw are penalized if their VP benefit is marginal.

Input‑aware: Consuming high‑VP inputs (e.g., rare ores/chemicals) counts as a cost, so the controller won’t cannibalize valuable intermediates unless it pays off.

Data‑driven and future‑proof

Works with your XML‑defined processes/resources—you merely supply the per‑process I/O kg/sol maps you already read at boot.  GitHub

Aligns with the regolith → ore → refined outputs loop that underpins mining & water recovery. If ore outputs have high VP (or their downstream products do), the controller naturally brings on the right upstream processes. 

Why this improves the gameplay loop

Turns exploration/mining into purpose‑driven production. Once scouts find ice/minerals and teams bring regolith/ores home, the settlement “does the right thing” without micromanagement—melting, cracking, extracting—until water/oxygen/food/parts are in the green. That matches the intended loop in your Minerals & Water‑Bound Ores and Economics docs.  GitHub
+1

Visible, tunable levers (targets & weights). You can expose targets in UI (or load from XML) so players set priorities (e.g., “rush oxygen tanks” vs “stock construction materials”).

Robustness, not brittleness. Hysteresis and min‑run remove the flip‑flop feel players often report when a system auto‑toggles too fast.

### DIFF
--- a/mars-sim-core/src/main/java/com/mars_sim/core/automation/ResourceProcessPlanner.java
+++ b/mars-sim-core/src/main/java/com/mars_sim/core/automation/ResourceProcessPlanner.java
@@ -1,0 +1,372 @@
+// package com.mars_sim.core.automation; // <-- adjust or leave default package if you prefer
+
+import java.util.*;
+import java.util.function.*;
+import java.util.stream.Collectors;
+
+/**
+ * ResourceProcessPlanner
+ *
+ * One-file, plug-and-play controller that automatically starts/stops resource processes
+ * to meet settlement stock targets using the project's Value Point (VP) economics.
+ *
+ * Design goals:
+ *  - Zero compile-time coupling to mars-sim classes (supply lambdas/adapters).
+ *  - Stable: hysteresis + minimum run time prevent flapping.
+ *  - Efficient: power-budget aware; penalizes high-power/low-yield processes.
+ *  - Extensible: you can add custom scoring weights without editing other files.
+ *
+ * Typical wiring:
+ *
+ *   var planner = ResourceProcessPlanner.<Settlement, Process>builder()
+ *       .listProcesses(settlement -> settlement.getProcesses()) // Iterable<Process>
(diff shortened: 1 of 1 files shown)


mokun:
ResourceProcessPlanner adopts a Builder pattern. The question is how to incorporate and replace what we currently have. It'll be a huge leap.

passingbyhelpingout1:
> ResourceProcessPlanner adopts a Builder pattern. The question is how to incorporate and replace what we currently have. It'll be a huge leap.

if you want to merge it feel free, ill untag draft, if not ill leave it for now just as a spare for later (a basic frmaework)

PR #1702, opened 2025-08-31
Replace Building Salvage with the improved Construction Mission

### LINKED ISSUE #1584: Create concept of MissionObjectives. Improve Mission Tool UI

### LINKED ISSUE #1699: Combine Salvaging to Construction Mission
## Problem

The current ```SalvageMission``` is very similar to the improved ConstructionMission but without the improvements. Sites now hold the phases of a construction as either construct or salvage.
This means there is less need for a dedicated salvage mission

## Solution

The propose solution is to remove the Salvage mission and all associated classes. The existing ConstructionMission can also perform the salvage actions with little change.
The required change will be a different outcome at the end of the mission depending upon construct/salvage.
To simplify the change the ConstructionStage needs to be aware of whether it is construct or salvage

bevans2000:
This PR removes all the explicit Building Salvage logic because it is now superseded by the new improved Construction Mission.
The number of file changed is larger than ideal but many are just method renames or change to return types.

Key highlights:

- Buildings are selectable for demolishing via the Settlement Map, giving the user control.
- Salvaging is now supported by same Construction logic; the explicit Salvage building logic is removed
- A construction site is defined with all the phases in the construction plan need Construction or Demolish
- ConstructionStage has more responsibility for managing the materials instead of being spread across numerous classes.
- ConstructionMission can action different results based on the construction flag.
- Update to the UI on Construction mission
- Correct problem of missing 7x7 foundation construction stage. ConstructionConfig ensures the prestages are correct

In addition, this allows the conversion to Mission Objectives to be completed.
close #1699 
close #1584


bevans2000:
> > now it is selected by the user from the Settlement Map instead of automatically which is better.
> 
> Guess in future, it'll need to be automatic at least for non-player/AI settlements.
Yes but it should be done differently, separate from the missions which was the case.

I've uncovered a worrying problem. Sometimes when triggering the demolishing of a building from the UI the simulation thread is mid pulse and there is a conflict. Such significant changes have to be done by the Simulation not the UI. So I think we need a similar concept to the SwingWorker that allows delegated execution of code. The execution would be synchronised with the clock.
It's just a case of having a central queue of ScheduledEventHandlers, reusing what we have already.

mokun:
> significant changes have to be done by the Simulation not the UI

Question: how do you switch the execution of a chunk of codes from one thread into another thread ?

passingbyhelpingout1:
> > significant changes have to be done by the Simulation not the UI
> 
> Question: how do you switch the execution of a chunk of codes from one thread into another thread ?

I doubt my seperate draft on multi threading would be any help.  Some form of cross shifting, but wouldnt that require current instructions to be held in some sort of short term memory buffer

bevans2000:
Agree my plan is to do it at weekend when I have time

bevans2000:
@mokun This is very annoying. The Unit tests have been changed such that after my merge from master they are failing. We need to make sure the master branch stays clean.

bevans2000:
I'll have to investigate this before you merge it. Arrgghhh

passingbyhelpingout1:
the patch - pull request i created seems to conflict with this file, because this file is what it is fixing, so you will have to merge this with the error in it, then apply my fix by merging it afterwards.  So faulty code through, then apply my code which fixes this issue and the whole thing is finalised ( hope that makes sense), once this first one is merged the conflict error on mine will go away then it can be merged in to fix the error

mokun:
> There is a randomness been introduced to the tests.

@bevans2000 

It's true that in MaintainBuildingEVATest and elsewhere, I use probability to determine if the meta task would result in generated a SettlementTask.

I'll have to think about how to rewrite it in order to avoid failing those unit tests.

Btw, I just fixed sometimes in the master branch. All good now.

passingbyhelpingout1:
> > pull request i created seems to conflict with this file, because this file is what it is fixing, so you will have to merge this with the error in it, then apply my fix by merging it afterwards
> 
> @passingbyhelpingout1
> 
> I'll take a look tomorrow on your patch on MaintainBuildingEVATest.

Youve fixed it so the patch may not be necessary anymore, ill leave it up for now, and ill check my 3d renderer submission to see if its gone

PR #1703, opened 2025-09-01
Update UnitManager.java

passingbyhelpingout1:
What this does (in plain English)

Process finished settlements as they complete (using ExecutorCompletionService) instead of in submit order. This lowers tail latency when one settlement is slow.

Never let one settlement crash the whole tick: we catch and log RuntimeException inside the settlement task, then keep going. Several real bugs previously propagated up through Future.get() and stalled the loop.  

Built‑in micro‑profiling: per‑settlement duration is measured; slow updates log a concise warning so you can see who’s heavy without a full profiler.

Barrier semantics preserved: the method still waits for all settlements each pulse (so game state remains consistent); we only change how we wait.

Why this matters: you’ve had hard exceptions thrown from UnitManager.runExecutor(...) because a settlement task failed (e.g., scientific‑study collaborator checks or mission anomalies), which percolated to Simulation.clockPulse(...) and the MasterClock task. This isolates that class of failure while giving you visibility into which settlement is slow.

### DIFF
--- a/mars-sim-core/src/main/java/com/mars_sim/core/UnitManager.java
+++ b/mars-sim-core/src/main/java/com/mars_sim/core/UnitManager.java
@@ -17,6 +17,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 
 import com.mars_sim.core.building.Building;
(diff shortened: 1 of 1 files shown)


passingbyhelpingout1:
This might be a more subtle fix, at least while i work on the other one i have as a draft for the parallel execution, and hopefully it should solve some of the short term issues that have popped up on this front.  If its not useful just close it off and all is well

mars-sim-core/src/main/java/com/mars_sim/core/UnitManager.java
```diff
@@ -312,7 +326,12 @@ public int getCommanderID() {
 	@Override
 	public boolean timePassing(ClockPulse pulse) {
 		if (pulse.getElapsed() > 0) {
```
mokun, line 328: @bevans2000 

What would be the situation when pulse.getElapsed is zero or negative ?

bevans2000: It would never be true

bevans2000: Generally I would not make any changes to the Unit manager directly on the master branch without significant testing as the impact of it breaking is very large

passingbyhelpingout1: Ok closing off for now

PR #1707, opened 2025-09-01
Update Moon.java

passingbyhelpingout1:
Below are safe, interop‑friendly improvements you can apply to com.mars_sim.core.moon.Moon that won’t break other subsystems. I designed these changes to be additive, immutable, thread‑safe, and unit‑explicit, so they play nicely with your multi‑threaded sim loop and mission scheduling (see the concurrent time‑stepping patterns visible in issue #1356’s stack trace).  GitHub

They also avoid hard dependencies on global clocks or Mars‑specific classes, making the type easy to use across modules and content pipelines (the repo has a content/ directory and multiple modules).  GitHub

What’s improved (without breaking other parts)

Immutability & Thread‑safety: All fields are final and validated; the object is Serializable. This avoids accidental mutation across the concurrent sim tickers.  GitHub

Unit clarity: Every quantity declares SI units (meters, seconds, radians, kg). No more ambiguous doubles.

Pure math API: Orbital position/angles computed as pure functions that accept an Instant epoch + Instant when — no global clock coupling.

Precomputed derived values: Mean motion (rad/s) is precomputed to reduce per‑tick cost.

Value semantics: equals/hashCode keyed to name (unique identity), plus a concise toString().

Builder: Fluent Builder for clean construction from configs (e.g., YAML/JSON loaders in your content pipeline) while keeping the core immutable.  GitHub

Input validation: Guardrails (e.g., clamp eccentricity to [0, 1), positive period). Defensive Objects.requireNonNull.

Additive API: The class stays self‑contained; there are no removals of prior public members. (If your current Moon exposes extra getters/ctors, keep them and delegate to the builder internally — easy to do if needed.)

Why these changes are safe with the rest of the sim

No global coupling: The class doesn’t pull in MasterClock or sim singletons, so it avoids hidden threading/cycle hazards typical in time‑stepped systems (your stack traces show parallel tasks via ThreadPoolExecutor).  GitHub

Configuration friendly: The Builder maps cleanly from external data sources (the repo layout indicates a content/ folder used to seed data), and the class remains a plain value with clear units.  GitHub

Additive API: You can keep older constructors/getters if they exist and implement them by delegating to the new internals shown. No removals are necessary.

Optional follow‑ups (all backward‑compatible)

Convenience factories: Moon phobos = Moon.builder("Phobos")...build(); (define constants in a separate Moons utility class so Moon stays general).

Caching hooks: If you call positionAt() at high rates with the same when, consider a tiny, optional memoizer (per‑thread cache) — but only if profiling says so.

Inter‑module bridges: If you have a global Mars/Master clock, add a trivial overload:

public double[] positionAt(MarsClock clock, Instant epoch) {
    return positionAt(epoch, clock.getInstant());
}

Implement it in a separate adapter to avoid a compile‑time dependency here.

Validation tests: Add unit tests that assert invariants and numerics (angles normalized, eccentricity clamped, mean motion derived).

### DIFF
--- a/mars-sim-core/src/main/java/com/mars_sim/core/moon/Moon.java
+++ b/mars-sim-core/src/main/java/com/mars_sim/core/moon/Moon.java
@@ -16,67 +16,84 @@ import java.util.Objects;
  *   <li>Time: seconds (s)</li>
  * </ul>
  *
- * <p><b>Design (interop‑friendly):</b>
+ * <p><b>Design goals</b>:
  * <ul>
- *   <li>Mutable POJO (no-arg ctor + setters) to avoid breaking legacy callers</li>
- *   <li>Validation on writes + angle normalization helpers</li>
- *   <li>Pure orbital math utilities; no global clocks/singletons</li>
- *   <li>Builder + copy/with methods for modern/immutable-style callers</li>
- *   <li>Alias getters/setters to keep ambiguous legacy names working</li>
+ *   <li>Mutable POJO (no-arg constructor + setters) for existing loaders/serializers.</li>
+ *   <li>Validation on write and angle normalization to avoid NaN/Inf propagation.</li>
+ *   <li>Pure orbital helpers with no global singletons or clocks.</li>
+ *   <li>Builder and copy/with helpers to support immutable-style use without breaking legacy code.</li>
+ *   <li>Alias getters/setters (e.g., {@code getRadius()}, {@code setPeriod(double)}) to preserve older call sites.</li>
  * </ul>
  */
 public class Moon implements Serializable {
 
+    /** Serialization ID for compatibility. */
     private static final long serialVersionUID = 1L;
 
-    // ----------------------------
-    // Identity & bulk properties
-    // ----------------------------
+    // ---------------------------------------------------------------------
+    // Identity & physical properties
+    // ---------------------------------------------------------------------
 
     /** Unique name (e.g., "Phobos", "Deimos"). */
-    private volatile String name;
+    private String name;
 
-    /** Physical mean radius [m]. For irregular bodies, use the accepted mean radius. */
-    private volatile double bodyRadiusMeters;
+    /** Physical mean radius (meters). For irregular bodies, use accepted mean radius. */
+    private double bodyRadiusMeters;
 
-    /** Mass [kg]. */
-    private volatile double bodyMassKg;
+    /** Mass (kilograms). */
+    private double bodyMassKg;
 
-    // ----------------------------
-    // Keplerian elements (parent-body equatorial frame)
-    // ----------------------------
+    // ---------------------------------------------------------------------
+    // Keplerian elements (parent-body equatorial frame, SI units)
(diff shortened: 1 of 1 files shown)


passingbyhelpingout1:
ill fx the conflict on this tomorrow with the Ci check, looks easy enough

bevans2000:
What problem is this solving? The whole Moon.java seems to have been changed

mars-sim-core/src/main/java/com/mars_sim/core/moon/Moon.java
```diff
@@ -1,28 +1,260 @@
...
+    private final double inclinationRad;          // radians (i)
+    private final double longAscNodeRad;          // radians (Ω)
+    private final double argPeriapsisRad;         // radians (ω)
+    private final double meanLongitudeAtEpochRad; // radians (L0) = M0 + ϖ
+    private final double siderealPeriod;          // seconds (T)
+
+    // ---- Derived, cached ----
+    private final double meanMotion; // radians per second, n = 2π/T
+
+    private Moon(Builder b) {
```
mokun: Good that you see this class can be tweaked with a Builder pattern to simulate any moons. 

@bevans2000 

I'll merge this and play with it. 

We have more or less defined the orbit of Mars around the sun but it can really use some improvement by defining more orbital mechanic parameters to further figure out how to go between Mars and Earth.

Here it reminds me we can do it better.

Plus we could even pinpoint where the two moons of Mars (namely Deimos and Phobos) are at and set up a mineral base on them and excavate raw resources, etc. 

For sure our Astro package may already have similar ability to do no less but is harder to understand. But I'd like to start from scratch and use this to put together more pieces until we can apply rocket equations and estimate payloads to be delivered to Mars

bevans2000: I think it's failing to build at the moment so don't merge it

passingbyhelpingout1:
Ok this is good to go for merging.  Codefactor and java CI / Build issues all cleared.  no conflicts with base branch

mars-sim-core/src/main/java/com/mars_sim/core/moon/Moon.java
mokun: @bevans2000 

I'll accept this if it's okay with you.

mars-sim-core/src/main/java/com/mars_sim/core/moon/Moon.java
```diff
@@ -1,28 +1,762 @@
-/*
```
mokun, line 1: Why deleting the date and author tag ?

mokun: This class file is a low-hanging fruit for sure since it's empty and contribute nothing to the simulation engine yet. Whatever our imagination will take us next.

I'll accept this PR now. 

But in future, please do keep the date and author tag. See to it that it's our convention to have a tag like this for each class file with source code. Thanks !

PR #1708, opened 2025-09-02
Update LunarProject.java

passingbyhelpingout1:
Why these changes are safe with the rest of the codebase

Constructor compatibility: The original package‑private 3‑arg constructor is preserved verbatim, so any existing factories/builders work unchanged. The new 4‑arg constructor is additive only.  GitHub

Field visibility & semantics: The logger field remains public static final (though non‑idiomatic) to avoid breaking any external references. getParticipants() still returns the live set, preserving prior mutability semantics.

Types tightened without binary breakage: Adding generics to fields/returns doesn’t alter the bytecode signatures due to type erasure; it only improves compile‑time safety. Callers compiled with raw types continue to work.  GitHub

addTopic() type matches config: ScienceConfig#getATopic(ScienceType) returns a String; changing topics to List<String> aligns the types and prevents accidental insertion of non‑string elements.  GitHub

Bug fix in capacity check: The corrected predicate in canAddParticipants() aligns with the method name and intended behavior; code that previously used it would have behaved inversely. This is a correctness fix.  GitHub

**This inital pull request meets all the checks and is ready to merge if it is accepted - intend to revisit later but the new additions seem suitable for now**

### DIFF
--- a/mars-sim-core/src/main/java/com/mars_sim/core/moon/project/LunarProject.java
+++ b/mars-sim-core/src/main/java/com/mars_sim/core/moon/project/LunarProject.java
@@ -1,14 +1,16 @@
-/*
+/* 
  * Mars Simulation Project
  * LunarProject.java
- * @date 2024-02-15
+ * @date 2024-02-15 (revised)
  * @author Manny Kung
  */
 package com.mars_sim.core.moon.project;
 
 import java.io.Serializable;
+import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Objects;
 import java.util.Set;
 
 import com.mars_sim.core.SimulationConfig;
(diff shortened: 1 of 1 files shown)


mars-sim-core/src/main/java/com/mars_sim/core/moon/project/LunarProject.java
```diff
@@ -17,69 +19,145 @@
 import com.mars_sim.core.science.ScienceType;
 import com.mars_sim.core.tool.RandomUtil;
 
+/**
+ * Represents a lunar research/engineering project in the simulation.
+ * <p>Thread-safety: not thread-safe; callers should synchronize externally if needed.</p>
+ */
 public class LunarProject implements Serializable {
```
mokun, line 26: @bevans2000 

The changes are harmful on this. I'll accept the pull.

passingbyhelpingout1: > @bevans2000
> 
> The changes are harmful on this. I'll accept the pull.

if it wasnt just a typo for "not harmful" if they are indeed harmful, let me know what parts need fixing and ill get it sorted, if it was just a typo well 👍

mokun: it looks harmless. The change is not too dramatic.

PR #1710, opened 2025-09-02
Create SpaceWeatherEvent.java

passingbyhelpingout1:
Space weather alerts & shelter drills

Why it fits: Radiation modeling already distinguishes SEP vs GCR and curtails EVA during SEP; we add proactive alerts, a small scheduler/generator, and a UI panel to make it visible and actionable.  GitHub

What it adds: early warnings, morale effects (optional), and “Shelter Drill” hooks—lightweight, no behavior changes unless you subscribe to the service.

### DIFF
--- a/mars-sim-ui/src/main/java/com/mars_sim/ui/swing/panel/SpaceWeatherPanel.java
+++ b/mars-sim-ui/src/main/java/com/mars_sim/ui/swing/panel/SpaceWeatherPanel.java
@@ -1,0 +1,59 @@
+package com.mars_sim.ui.swing.panel;
+
+import com.mars_sim.core.environment.spaceweather.SpaceWeatherEvent;
+import com.mars_sim.core.environment.spaceweather.SpaceWeatherListener;
+import com.mars_sim.core.environment.spaceweather.SpaceWeatherService;
+
+import javax.swing.*;
+import java.awt.*;
+import java.time.Duration;
+import java.time.Instant;
+import java.util.Timer;
+import java.util.TimerTask;
+
+/**
+ * Minimal Swing panel that shows the current space weather state.
+ * You can add this to CommanderWindow/Monitor tool without refactoring.
+ */
+public class SpaceWeatherPanel extends JPanel {
+
+    private final JLabel status = new JLabel("—");
+    private final JProgressBar riskBar = new JProgressBar(0, 100);
+    private final SpaceWeatherService svc;
+
+    public SpaceWeatherPanel(SpaceWeatherService svc) {
+        super(new BorderLayout(8, 8));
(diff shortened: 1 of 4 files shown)


passingbyhelpingout1:
This one is a little bit more ambitious, but basically adds a base layer of which to build upon space weather, i can expand later for surface side events and the like.  for now just a basic framework and would add another solid layer at least IMO to the game

mokun:
Yay it doesn't hurt to add space weather event.

Let's see if you can integrate it and track it using our existing HistoricalEvent class. 

A number of other types of events are extending HistoricalEvent.

If you can't, I'll try to incorporate it manually once you're done tweaking it.

I suppose in future, the dust storm monitoring will be a part of the space weather service as well.

Your predecessor once suggest adding tracking dust storm with pull requests. 

However, it looks like all the codes regarding that have been deleted.

bevans2000:
Also how does a weather event affect other entities in the simulation

passingbyhelpingout1:
> Also how does a weather event affect other entities in the simulation

For now it would be around dust storms, increasing dust on solar panels, leading to creating a job to clean them, perhaps some of the dust can be converted into regolith, which could then be used in construction.  That is the longer term plan at least and fully in line with what is generally accepted science and activities for a marsian colony

bevans2000:
So you need to change it so it is not using Instant now because that doesn't work.
Really if we are introducing a new concept we need to put it in place otherwise we are just collecting unused bits of code.
The PRs would be better if they deliver a complete new feature. For bigger changes it doesn't have to solve everything but needs to be used when we run it. And needs unit tests

mars-sim-ui/src/main/java/com/mars_sim/ui/swing/panel/SpaceWeatherPanel.java
```diff
@@ -0,0 +1,59 @@
...
+            @Override public void onSpaceWeather(SpaceWeatherEvent e) {
+                SwingUtilities.invokeLater(() -> update(e));
+            }
+        });
+
+        // Optional: drive service locally if not wired to master clock yet.
+        Timer timer = new Timer("SpaceWeatherPanelSim", true);
+        timer.scheduleAtFixedRate(new TimerTask() {
+            private Instant t = Instant.now();
+            @Override public void run() {
```
mokun, line 44: @passingbyhelpingout1 

Can you switch to receiving a ClockPulse instance in a timePassing() method ? 

This is how we hook up a new class instance into receiving the MarsTime from a clock pulse ? 

There are plenty of examples of the use of timePassing() in all unit related class that you can wire into, instead of using Timer here to do the job.

PR #1711, opened 2025-09-02
Create SolarPanelDustComponent.java

passingbyhelpingout1:
Feature 1 (with code): Dust on solar arrays + cleaning task Design snapshot

New helper SolarPanelDustComponent (core, pure logic): updates panel dust based on optical depth & wind; exposes an efficiency factor.

Integrate in the power function: multiply raw solar power by dustEfficiency.

New Task CleanSolarPanels + CleanSolarPanelsMeta: When dust > threshold and weather safe, a maintenance‑type task schedules EVA cleaning (time ~ exposed area).

Optional XML: allow a <power-generation type="solar" panel-area="…"> attribute in buildings.xml (kept backward‑compatible; if absent, assume a default area). XML override lives under ~/.mars-sim/xml, as documented. 

Dust storms that foul solar arrays + a “Clean Solar Panels” maintenance task

Weather metrics already modeled include optical depth, wind speed and solar irradiance; these are perfect drivers for dust deposition and self‑cleaning. The Building Functions include POWER_GENERATION, and Tasks/MetaTasks are the AI entry points to make settlers do the cleaning. So we can attach a dust model to solar power buildings, reduce output when dusty, and surface a new task that schedules EVA cleaning when safe.

### DIFF
--- a/mars-sim-core/src/main/java/com/mars_sim/core/power/SolarPanelDustComponent.java
+++ b/mars-sim-core/src/main/java/com/mars_sim/core/power/SolarPanelDustComponent.java
@@ -1,0 +1,113 @@
+/*
+ *  Mars Simulation Project – Solar panel dust model
+ *  Copyright (C) 2025
+ *
+ *  This program is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation, version 3.
+ *
+ *  This program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU General Public License for more details.
+ */
+package org.mars_sim.msp.core.power;
+
+import java.io.Serializable;
+import java.util.Objects;
+
+/**
+ * Tracks dust deposition on a building's exposed solar panels and provides an efficiency factor (0..1).
+ * Drive it each tick with current optical depth (tau), wind speed (m/s), and dt (millisols or seconds).
+ *
+ * Assumptions (tunable):
+ *  - Deposition scales with atmospheric optical depth (tau) during daylight.
+ *  - Strong winds provide partial self-cleaning (gust threshold).
+ *  - Manual cleaning reduces dust at a fixed rate per m^2 cleaned.
+ */
+public class SolarPanelDustComponent implements Serializable {
+
+    private static final long serialVersionUID = 1L;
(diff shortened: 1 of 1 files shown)


bevans2000:
This cleaning is a good idea. This should be combined with your clean panel task so it is delivered as one complete piece of work instead of individual files?

passingbyhelpingout1:
> This cleaning is a good idea. This should be combined with your clean panel task so it is delivered as one complete piece of work instead of individual files?

Wanted to make sure given there seems to be some resistance around here to major changes in big batch files taht the underlying solar panels and power generation got accepted first and then if they got through ( and if you had any changes to events or anything else) that at least then these could be added later on.  More keeping it segmented so that weather , modules , events and jobs are intergrated seperate, if you want to approve i think the 4 that form this batch then they should all be able to  go together.  And then i can iterate further

bevans2000:
I'm just applying the usual coding approach. Creating isolated classes that can not be used in context makes it difficult. I would prefer to group the changes into PR that are focused on a single well defined.

passingbyhelpingout1:
> I'm just applying the usual coding approach. Creating isolated classes that can not be used in context makes it difficult. I would prefer to group the changes into PR that are focused on a single well defined.

Fair point, ill try and make sure in future that there is some linking of various file contributions.  The plan was to begin to link some of the scaffolds together once they had been approved.  An iterative approach over time.  The next batch are a new thing, 4 files 3 new and one simulation.java tweak to incorporate the idea around the 3 spaceweather files.  Look them over when you get a chance

bevans2000:
It can be hard to understand what the individual files provide without seeing them linked together into the simulation. For the dust modelling I would also create an overarching Issue where you set out what problem/feature you want to solve. Gives context to the change

mokun:
> a new task that schedules EVA cleaning when safe.

Yes it'll be interesting in future to model the scheduling of an solar panel cleaning SettlementTask.

mokun:
> The next batch are a new thing, 4 files 3 new

I can see you're breaking the work down into manageable chuck and a few PRs that comes after this one.

mokun:
> Creating isolated classes that can not be used in context makes it difficult.

Yes our preference is to have may be at least two related class files so that we can have an overview on how it would impact the codebase.

But this time I'll merge, since you have already generated this class file in the next few PRs.

mokun:
> The unit tests are failing now

That's a separate issue.

It said 

` ExperienceImpactTest.testApplyBasicSkill:53 Unchanged skill expected:<174.0> but was:<249.0>`

We have these from time to time.

Re-running it now.

PR #1715, opened 2025-09-03
Update SettlementMapPanel.java

passingbyhelpingout1:
One high‑impact coding fix:
Stop the Settlement Map zoom memory leak by coalescing zoom events, removing listener leaks, and caching scaled SVG renders.

Why this matters
There’s an open bug where dragging the zoom slider (or using the mouse wheel) in the Settlement Map rapidly drives up heap usage and it doesn’t recover even after closing the map. The report points at the Swing client and mentions recent use of higher‑fidelity SVGs (e.g., bunk‑bed art) as a likely contributor. Fixing this will make the UI feel snappier and far more stable in long sessions.  GitHub
+1

Context to ground the change

The “JavaUI” label on the issue indicates the Swing client path (not the JavaFX builds). The README also notes the official release doesn’t require JavaFX, so improvements belong in the Swing UI module.  GitHub
+1

The repo layout shows a mars-sim-ui module where the Swing panels live (e.g., SettlementMapPanel, SettlementTransparentPanel).  GitHub

What to change (summary)

Coalesce zoom changes so you don’t re-render on every slider tick or wheel notch.

Ensure listeners are not leaked (add/remove exactly once; tear them down on panel disposal).

Add an LRU cache of scaled rasterized icons (SVG → BufferedImage) keyed by resource and scale factor, using SoftReference so the GC can reclaim under pressure.

Together, these eliminate the hot path that allocates new images and listeners on every tiny zoom delta—precisely the pattern that produces the rising “Max memory” line described in the issue.

How to verify (quick manual repro + acceptance)

Start the Swing client, open the Settlement Map, and drag the zoom slider up and down for 10–20 seconds. Before the fix, “Max memory” climbs into the hundreds of MBs/GBs; after the fix, it should plateau and then fall back within a few seconds of idling.  GitHub

Close the Settlement Map window/dialog. After a short pause (or after a GC), heap usage should return close to its baseline instead of staying high.

Optional: programmatic sanity check—simulate ~1,000 zoom changes, call System.gc() and sample MemoryMXBean.getHeapMemoryUsage().getUsed(); ensure it remains near steady state.

Why this is the “one thing” to do

It targets a public, high‑priority pain point users will hit early (zooming the map).  GitHub

It’s self‑contained (pure Swing/UI), aligns with the project’s current official UI (Swing, not JavaFX), and avoids architectural churn.  GitHub

It leverages the project’s current use of SVG assets by caching rather than recreating them, which is the standard remedy for zoom‑driven leaks. The SVG work referenced (e.g., bunk‑bed art) will benefit directly.

### DIFF
--- a/mars-sim-ui/src/main/java/com/mars_sim/ui/swing/tool/settlement/SettlementMapPanel.java
+++ b/mars-sim-ui/src/main/java/com/mars_sim/ui/swing/tool/settlement/SettlementMapPanel.java
@@ -17,9 +17,12 @@ import java.awt.RenderingHints;
 import java.awt.event.MouseAdapter;
 import java.awt.event.MouseEvent;
 import java.awt.event.MouseMotionAdapter;
+import java.awt.image.BufferedImage;
+import java.lang.ref.SoftReference;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.LinkedHashMap;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
@@ -29,6 +32,7 @@ import java.util.Set;
 
 import javax.swing.JPanel;
 import javax.swing.SwingUtilities;
+import javax.swing.Timer;
 
 import com.mars_sim.core.CollectionUtils;
 import com.mars_sim.core.Unit;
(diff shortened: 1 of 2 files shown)


mars-sim-ui/src/main/java/com/mars_sim/ui/swing/tool/settlement/SettlementMapPanel.java
```diff
@@ -172,6 +205,10 @@ public SettlementMapPanel(MainDesktopPane desktop, final SettlementWindow settle
 		selectedBuilding = new HashMap<>();
 		selectedPerson = new HashMap<>();
 		selectedRobot = new HashMap<>();
+
+		// New: throttle zoom changes to avoid flood of repaints/rasterizations
+		zoomCoalesceTimer = new Timer(50, e -> applyPendingScale());
```
mokun, line 210: Yes creating a time delay to avoid creating new buffered images by every tiny change of scale is a very good strategy.

Issue #1670, opened 2025-08-23
Salvage Building should support reusing a frame


bevans2000:
**Is your feature request related to a problem? Please describe.**
The Salvage Building Mission current demolishes the whole Building which is expensive and time consuming.
The user should be given the ability to repurpose a Building instead of destroying the whole location. This would mean the foundations could be re-used for the new Building without starting from scratch,

**Describe the solution you'd like**
This is a follow on from the changes in #1642.
The change would be to convert the selected ```Building``` to a ```ConstructionSite``` based on the user's trigger. The new site would have the replacement Building as the target as well as being tagged as Salvage.
Using the new approach the Salvage Mission will select this Site and start the Salvaging. Salvaging will continue through the ConstructionStagesInfo until a common one is reached between the old & new Building. At this point the site will be changed to be construction and hence ConstructionMission will take over.

This will include a new user pop-up offering the user all potential new buildings that share a common Construction Stage with the existing one.

**Java classes involved**
```SalvageMission``` & ```Constructionmanager```

**Additional context**
This might include reworking the Salvage Building Task to be closer to the Construction equivalent.

PR #1716, opened 2025-09-03
Update SettlementMapPanel.java

passingbyhelpingout1:
One high‑impact, code‑level improvement:
Fix the open memory leak when dragging the Settlement Map zoom slider by (a) coalescing slider events and (b) bounding/cleaning the map‑image cache.

Why this is worth doing

There is an open bug: “Memory leak when using zoom slider in Settlement Map” (#1670; labeled “JavaUI / Bug”). Fixing it improves stability and responsiveness for everyone using the Swing client.  GitHub

The project’s official releases use the Swing UI (JavaFX isn’t required), so this targets the mainline UI where players feel it.

Verify the fix (quick steps)

Launch the Swing client → open Settlement Map.

Use VisualVM or JFR to watch heap & retained instances.

Drag the zoom slider back and forth for ~30 seconds.

Before: retained BufferedImage steadily increases; heap rises after GCs.

After: retained images plateau; heap returns to baseline after GC cycles.

Notes & sources

Project path & class name (org.mars_sim.msp.ui.swing.tool.settlement.SettlementMapPanel) are long‑standing; they show up in old CI logs too.  SourceForge

Swing is the official UI path (JavaFX not required), and Java 21 is the supported JDK, so the approach avoids new deps and sticks to Swing idioms.  GitHub

Zoom slider exists in the Settlement Map tool, per the user guide.

I kept your existing architecture (layered rendering, throttled setScale, listener lifecycle, and ScaledIconCache) and added:

Attach/detach Zoom JSlider helpers using coalesced updates (safe to call from your transparent panel or any toolstrip).

A bounded, GC‑friendly offscreen base cache (BaseBufferCache) for the background layer (which is the heavy, scale‑dependent part). The cache is small, uses SoftReference, and is cleared on scale/rotation/pan changes to avoid stale images.

Extra lifecycle hygiene in removeNotify() (stop timer, clear caches, detach slider) in addition to your destroy().

The background layer is cached; dynamic layers (day/night, buildings, vehicles, people, robots) are still drawn live so the map stays correct while the simulation advances.

### DIFF
--- a/mars-sim-ui/src/main/java/com/mars_sim/ui/swing/tool/settlement/SettlementMapPanel.java
+++ b/mars-sim-ui/src/main/java/com/mars_sim/ui/swing/tool/settlement/SettlementMapPanel.java
@@ -2,7 +2,7 @@
  * Mars Simulation Project
  * SettlementMapPanel.java
  * @date 2025-08-01
- * @author Scott Davis
+ * author Scott Davis
  */
 package com.mars_sim.ui.swing.tool.settlement;
 
@@ -29,10 +29,13 @@ import java.util.List;
 import java.util.Map;
 import java.util.Properties;
 import java.util.Set;
+import java.util.function.Supplier;
 
 import javax.swing.JPanel;
 import javax.swing.SwingUtilities;
 import javax.swing.Timer;
+import javax.swing.JSlider;
+import javax.swing.event.ChangeListener;
 
 import com.mars_sim.core.CollectionUtils;
 import com.mars_sim.core.Unit;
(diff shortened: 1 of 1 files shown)


passingbyhelpingout1:
Ok if this works as i hope it should, it should be good for merging if it meets with everyones final approval

mars-sim-ui/src/main/java/com/mars_sim/ui/swing/tool/settlement/SettlementMapPanel.java
```diff
@@ -1091,9 +1156,43 @@ public void paintComponent(Graphics g) {
...
+								}
+								return img;
+							}
+					);
+					g2d.drawImage(base, 0, 0, null);
+				}
+
+				// 2) Draw remaining live layers (dynamic)
 				for (SettlementMapLayer layer : mapLayers) {
+					if (layer instanceof BackgroundTileMapLayer) continue; // already drawn via cache
```
mokun, line 1198: @passingbyhelpingout1 @bevans2000,

Last time I checked, the BackgroundTileMapLayer seems to be the culprit for the memory leak.

I tested it by turning off this background tile layer. So the background of the settlement map became black. 

Then I scrolled up and down the zoom slider, the memory went up and down very minimally.

So it must be somehow when rendering the BackgroundTileMapLayer, the memory was leaking. 

So our next task may just focus on fixing BackgroundTileMapLayer.

passingbyhelpingout1: ok ill close this, as i may have found a solution by fixing backgroundtilemaplayer, it supercedes this one.  And the new one modifies both files in one go

Issue #1662, opened 2025-08-19
Sleep table update causing sleep time to go 0 millisols


mokun:
**Describe the bug**
- When speeding up to 54,985x or above, the Sleep & Exercise table does not display correct value of sleep time.

**Discuss any potential problems you have and have not ruled out**
- `TabPanelHealth`'s `SleepExerciseTableModel::update` override the sleepTime and exerciseTime map.
- This cause the table to refresh itself often. At high speed, they don't sync up well.
- This suddenly leaves the first 3 or 4 rows of data with zero sleep time.

```java
		public void update(CircadianClock circadian) {
			sleepTime = circadian.getSleepHistory();
			exerciseTime = circadian.getExerciseHistory();
			
			Set<Integer> largestSet;
			if (sleepTime.size() > exerciseTime.size()) {
				largestSet = sleepTime.keySet();
			}
			else {
				largestSet = exerciseTime.keySet();
			}

			// Find the lowest sol day in the data
			solOffset = largestSet.stream()
					.mapToInt(v -> v)               
	                .min()                          
	                .orElse(Integer.MAX_VALUE);
			rowCount = largestSet.size();

			fireTableDataChanged();

```

**Screenshots**
- Currently, it only stores 7 days of data.
- On sol 51, the sleep table shows Gregory spent 432.2 millisols in sleep on on sol 45.
<img width="521" height="625" alt="Image" src="https://github.com/user-attachments/assets/efb9354f-817c-406c-8531-d6b7cbc12a56" />

- On sol 52, this sleep table got updated. it deleted all the hours on sol 45. It showed 0.0 millisols.
- It should either delete this sol row and not leaving it at zero hours.
<img width="531" height="635" alt="Image" src="https://github.com/user-attachments/assets/41fc2a1a-80e3-484a-ae1c-856313c2c58d" />

- At higher speed or time ratio, it gets worse. One can even see 3 or 4 rows of data with zero millisols.

**Specifications**
- pre-3.10.0 

**Additional**
- Need to look into why at high speed, the sleep time goes higher. 
   - It's possible that somehow the amount of fatigue got drifted higher than what it should be, thus causing the sleep time to increase in order to compensate it. It's not proven yet if this is the cause yet.

bevans2000:
Tasks should always consume a slice of the available pulse. Death is the only activity that in theory consumes everything

wonderinglostsoul44:
eep Sleep & Exercise rows from dropping to 0.0 millisols

[...]

PR #1717, opened 2025-09-04
Update TabPanelHealth.java

passingbyhelpingout1:
Make the Sleep & Exercise table model take a snapshot (not a live reference) and coalesce UI refreshes. This addresses open bug #1662 where rows intermittently show 0.0 millisols at high sim speeds because the table points at maps that are being updated while the UI re-renders.  GitHub

Why this helps

The issue describes SleepExerciseTableModel.update(...) reassigning sleepTime / exerciseTime from CircadianClock, then immediately recomputing solOffset, rowCount, and calling fireTableDataChanged(). At high time ratios, those maps change during rendering, so the model’s keys/row count become inconsistent (leading to “0.0 millisols” entries).  GitHub

Snapshotting into new, local maps on the EDT and coalescing table refreshes (e.g., 60–120 ms) removes the race and prevents UI thrash.

Notes on the change

No functional behavior change beyond correctness: we still show up to 7 recent sols; we simply drop off older rows instead of showing 0.0 placeholders (the bug report specifically shows rows becoming zero when they should be removed).  GitHub

We do not hold references to circadian.getSleepHistory() / getExerciseHistory(); we copy them to LinkedHashMap snapshots and render from those.

We drive updates through a small coalescing timer (100 ms is plenty). At 55k× sim speed, the UI would otherwise thrash with fireTableDataChanged() calls.

Be sure that getRowCount(), getColumnCount(), and getValueAt(...) read from these snapshot maps and rowCount.

How to verify (manual, quick)

Rebuild, run Swing UI, open the Health tab (Sleep & Exercise table).

Increase time ratio past 54,985× (per issue). Rows should no longer flip to 0.0 millisols; older rows drop off cleanly, newest rows stay stable.  GitHub

Watch EDT stalls and GC in VisualVM: update frequency will be dramatically lower vs. one fireTableDataChanged() per sim tick.

Why this is the “one thing” to do now

It fixes an open, reproducible bug that hurts UX at high speeds, with very small code changes and no new dependencies.  GitHub
+1

The pattern (snapshot → coalesce → fire on EDT) is reusable across other UI tables that stream sim data.

### DIFF
--- a/mars-sim-ui/src/main/java/com/mars_sim/ui/swing/unit_window/person/TabPanelHealth.java
+++ b/mars-sim-ui/src/main/java/com/mars_sim/ui/swing/unit_window/person/TabPanelHealth.java
@@ -14,6 +14,9 @@ import java.awt.event.MouseEvent;
 import java.text.DecimalFormat;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
+import java.util.LinkedHashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -27,6 +30,7 @@ import javax.swing.JScrollPane;
 import javax.swing.JTable;
 import javax.swing.JTextArea;
 import javax.swing.SwingConstants;
+import javax.swing.SwingUtilities;
 import javax.swing.table.AbstractTableModel;
 import javax.swing.table.DefaultTableCellRenderer;
 import javax.swing.table.TableColumnModel;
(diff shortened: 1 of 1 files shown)


passingbyhelpingout1:
Ok this should fix #1662 all going well, subject to approval and merging.

mars-sim-ui/src/main/java/com/mars_sim/ui/swing/unit_window/person/TabPanelHealth.java
```diff
@@ -999,54 +1012,77 @@ else if (columnIndex == 2) {
...
+
+			scheduleRefresh();
+		}
+
+		private void scheduleRefresh() {
+			if (SwingUtilities.isEventDispatchThread()) {
+				// Coalesce bursts (restart timer)
+				refreshCoalesce.restart();
+			} else {
+				SwingUtilities.invokeLater(() -> refreshCoalesce.restart());
```
mokun: @bevans2000 

Do you think TabPanelHealth should have already been on Swing EDT already ?

Do we need to use SwingUtilities.invokeLater() at all ?

bevans2000:
I haven't looked at this problem so not sure what's needed to fix it so don't feel in a opposition to comment. But I would be surprised if it needed any clever/deep Swing logic.

PR #1723, opened 2025-09-05

passingbyhelpingout1:
One high‑impact coding fix:
Stop the Settlement Map zoom memory leak by (a) removing duplicate mouse listeners when the map panel is closed/reopened and (b) switching the zoom pipeline to transform‑based rendering + a tiny debounce, instead of re‑rasterizing on every slider tick.

Why this is worth doing

There’s an open bug where “rolling the mouse wheel or dragging the zoom slider” makes heap usage climb and not come back down, even after closing the map. The report names SettlementTransparentPanel and SettlementMapPanel and notes listeners aren’t removed.  GitHub

The map uses Apache Batik to draw SVGs (build logs reference Batik classes and the exact file paths for SettlementMapPanel). Batik is great, but re‑building GVT trees or re‑rasterizing images repeatedly will balloon memory unless you reuse contexts and free caches.  SourceForge

What to change (one cohesive PR)
1) Make listeners idempotent and clean up on detach

In SettlementTransparentPanel (Swing), only attach listeners once, and remove them in removeNotify() (also handle HierarchyEvent.DISPLAYABILITY_CHANGED). This prevents retained references that keep the panel (and its bitmaps) alive after the window is closed.

### DIFF
--- a/mars-sim-ui/src/main/java/com/mars_sim/ui/swing/tool/settlement/SettlementMapPanel.java
+++ b/mars-sim-ui/src/main/java/com/mars_sim/ui/swing/tool/settlement/SettlementMapPanel.java
@@ -57,14 +57,13 @@ import com.mars_sim.ui.swing.UIConfig;
  * <p><b>What's new in this version</b> (memory‑safety & UX fixes):
  * <ul>
  *   <li><b>Zoom coalescing</b>: calls to {@link #setScale(double)} are throttled via a Swing Timer
- *       to prevent excessive re-rasterization while a slider is dragged.</li>
+ *       to prevent excessive re-rendering while a slider is dragged.</li>
  *   <li><b>Listener lifecycle</b>: mouse listeners are installed once and removed on dispose to
  *       avoid leaks when the panel is recreated.</li>
  *   <li><b>Graphics hygiene</b>: uses a child {@code Graphics2D} and disposes it after painting.</li>
  *   <li><b>Icon cache hook</b>: an LRU {@code ScaledIconCache} is exposed for layers that render
  *       scalable art to reuse rasterizations at the current scale.</li>
- *   <li><b>Tile cache cleanup</b>: calls {@link BackgroundTileMapLayer#dispose()} on remove to
- *       promptly free background tile images when the panel is detached.</li>
+ *   <li><b>Tile cache cleanup</b>: background tile images are released during panel destroy.</li>
  * </ul>
  * </p>
  */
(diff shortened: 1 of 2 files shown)


passingbyhelpingout1:
Ok this latest iterative fix / update is ready to go, passes all the checks with codefactor and java CI.  If this passes approval it may not fully fix the issues, but it does update the last appleid fix with the next stage that mokun mentioned the last one fixed some of the memory leak, but it may go deeper, so this one is the next stage approach to fixing the next component.  I can always get to work on the next stage if the issue goes deeper or elsewhere.

passingbyhelpingout1:
Some additional improvements that should further align and hopefully catch any follow on memory leaks that could potentially still be there after the inital committ.  preempting the next batch ( at least thats my hope)

mars-sim-ui/src/main/java/com/mars_sim/ui/swing/tool/settlement/SettlementMapPanel.java
```diff
@@ -1119,7 +1151,19 @@ public SettlementTransparentPanel getSettlementTransparentPanel() {
 	}
 
 	void update(ClockPulse pulse) {
-		settlementTransparentPanel.update(pulse);
+		// Clock pulses arrive on worker threads (MasterClock/ThreadPool). Swing must be updated on the EDT.
```
mokun, line 1154: Good point and the clock pulse is not on EDT. 

Either call SwingUtilities.invokeLater here or inside settlementTransparentPanel.update().

mars-sim-ui/src/main/java/com/mars_sim/ui/swing/tool/settlement/SettlementTransparentPanel.java
```diff
@@ -288,6 +294,16 @@ public Dimension getPreferredSize() {
 
         centerPanel.add(eastPane, BorderLayout.EAST);
 
+        // Patch 1: listen for mapPanel displayability change to detach listeners to avoid leaks
```
mokun, line 297: In future, it may be better to roll everything regarding mouse listener and zoom slider listener from SettlementTransparentPanel into SettlementMapPanel. 

Way too many coupling between them now. 

@bevans2000, is that possible ?

mokun:
I'll accept the PR and try it out.

mokun:
The memory leak still exists after the patch. Practically no change.

For now, in order to avoid the heap memory from going beyond 1 GB, my trick is that when I  zoom in, I won't go above 60x in scale. 

Beyond 60x, it balloons up to above 1 GB easily.

PR #1737, opened 2025-09-13
Create MarsZone

bevans2000:
This PR expands the msol offset of the Settlement into ```MarsZone```. This class represent 20 time zone around the planet each of 50 millisols.
There is a new unit test and the dependent classes like Event schedule have been updated.
The zone is also displayed in the weather tab of the Settlement unit window


mars-sim-ui/src/main/java/com/mars_sim/ui/swing/unit_window/structure/TabPanelWeather.java
```diff
@@ -211,80 +156,61 @@ private String getLongitudeString() {
...
 	        	solarIrradianceLabel.setText(StyleManager.DECIMAL_W_M2.format(solarIrradianceCache));
 	        }
 	    }
     }
-
-	/**
-     * Prepares object for garbage collection.
-     */
-    @Override
-    public void destroy() {
```
mokun, line 275: oh we aren't supposed to have destroy() ?

mars-sim-core/src/main/java/com/mars_sim/core/time/MarsZone.java
```diff
@@ -0,0 +1,95 @@
...
+        return offset;
+    }
+
+    /**
+	 * Gets the Mars time zone time offset for a given point on the surface.
+	 * 
+	 * @param point
+	 * @return
+	 */
+	public static MarsZone getMarsZone(Coordinates point) {
```
mokun, line 68: Great that we have this class now for tracking time zone.

mars-sim-ui/src/main/java/com/mars_sim/ui/swing/unit_window/structure/TabPanelWeather.java
bevans2000: The use of destroy is a bad anti-pattern. It our case we only need destroy if there is a listener that needs deregistering. Just nullify attributes serves no purpose. The base class have a default implementation.

bevans2000: If the destroy method only nullify variables I generally remove them

Issue #1698, opened 2025-08-30
Facilitate conversion between two reference frames (building versus settlement)


mokun:
**Describe the bug**
- Currently, if a building is centered at (10, 0), then confusions arise from time to time between a bed location or an activity spot (1, -1) within this building frame, versus its true location (11, -1) under the settlement positioning system.
- It's very prone to error if extra care is not provided to convert a location from one frame into another.

**Expected and unexpected behavior**
- Create a new class called `SettlementPosition` if a xloc and yloc are under settlement positioning system. 
- This will go in contrast with `LocationPosition`, which will be for a bed or an activity position defined in buildings.xml or BuildingSpecs. 
- Provide a new utility class (or may be using the old `LocalAreaUtil`) for explicitly conversing between the location between the two reference frame. 

**Java classes involved**
-  `LocationPosition`, `LocalAreaUtil`

**Stacktrace**
```
01-Adir-02:421.450 (Info) Function : [Teslapolis] TeslaBot-MR - Slot 2 @ Lander Hab 1 was claimed.
01-Adir-02:421.450 (Info) Task : [Teslapolis] TeslaBot-MR - Walking toward Lander Hab 1.
01-Adir-02:421.450 (Info) Task : [Teslapolis] TeslaBot-MR - Claimed the spot. Walking toward Lander Hab 1.
01-Adir-02:421.450 (Info) Task : [Teslapolis] TeslaBot-MR - Already at (-6.0, 0.0) in Lander Hab 1.
01-Adir-02:421.450 (Info) Function : [Teslapolis] TeslaBot-MR - Mgt 1 @ Lander Hab 1 was claimed.
01-Adir-02:421.450 (Severe) WalkingSteps : [Teslapolis] TeslaBot-MR - Invalid Robot start location at (-6.0, 0.0), not within Lander Hab 1 @ (0.0, 0.0)  Object: 9.0 W x 9.0 L at facing 0.0 deg.
01-Adir-02:421.450 (Info) Task : [Teslapolis] TeslaBot-MR - Walking toward Lander Hab 1.
```

**Specifications**
- pre-3.10.0

mokun:
It looks like the building local activity spots are being converted into settlement location in `Function` : 

```java
	/**
	 * Create a set of Activity Spots from a set of LocalPositions. The Activity spots are created
	 * into the global Settlement coordinate frame.
	 * 
	 * @param set Source positions
	 * @param owner     Building that owns this function, provide reference point.

	 */
[...]

Issue #1619, opened 2025-07-20
BuildingSpec contains parts of FunctionSpec


bevans2000:
## Problem
The ```Function``` subclasses use a ```FunctionSpec``` to construct the instance with the appropriate standard attributes, e.g. Activity Spots and attributes. Some Functions need extra structured data out of the XML file, e.g. Storage or Vehicle Maintenace. This extra configuration data is provided by the ```BuildingConfig``` via explicit methods by a lookup for the ```BuildingSpec```.
This provides a close coupling between the BuildingSpec and the specific Functions which is not advisable. 
Secondly it means the some of the Function need a reference to the BuildingConfig instance.
Following the common Instance -> Definition pattern the Definition should provide all the details of the Instance

## Solution
A better design would be that each Function can be created solely from the details of the ```FunctionSpec``` without a need for anything else. The correct design would be to use specific ```FunctionSpec``` subclasses where extra properties are needed. The Functions could then be constructed solely from their appropriate FunctionSpec giving a cleaner design and reducing coupling.
The BuildingConfig could use an AbstractFactoryPattern to create the appropriate FunctionSpec.
Issue #1562 has provided an example for the Manufacture function.

mokun:
I didn't see improvement until now. 

> A better design would be that each Function can be created solely from the details of the FunctionSpec without a need for anything else.

Yes to add on top of that, each FunctionSpec only need to be created once, and BuildingSpec too.

Currently at the start of the sim, it does seem like a waste to iterate over and over all creating those `Function` subclass for each building. 

So if there's a 100x Compact Quarters buildings in a settlement template, instead of creating 100x Building instance, may be we just need to create one and only one FunctionSpec and BuildingSpec and go from there.

I suppose what you meant is more than that.

So would you like to optimize that before releasing 3.10.0 ?

bevans2000:
I think I need to complete the Construction redesign first. This isn't actually causing a problem in the simulation so it can wait.

PR #1738, opened 2025-09-13
Convert Vehicle maintenance, EVA and medical to use Relative position 

bevans2000:
This PR applies 2 related changes (#1698 & #1619) to the ```VehicleMaintenance``` & ```Medical Function``` classes:

- Replace the use of LocalPosition in configuration classes with RelativePosition to avoid incorrect use.
- Remove Function specific properties from the base configuration classes

In addition the following modification we a consequence:
- VehicleGarage is removed as it is redundant; impacts numerous callers
- Single Generics based approach to tracking parking  using the new```VehcileLocation<T>```
- Parking locations are named
- Beds are now named
- The location of EVA ingress and engress are now explicit fields instead of relying on untyped Properties.


bevans2000:
@mokun Once I merge in your changes I'll mark this ready for merging.
I like the outcome of these change as I think it clears up logic.

PR #1742, opened 2025-09-14
Added one more rover name

Urwumpe:
Add some more rover and settlement names from German language and culture.


Urwumpe:
Looked odd, something went wrong there.  I repeat the process, I fear I had some bad local commits in it.

PR #1753, opened 2025-09-20
Make Coordinate parsing tolerant of language binding

### LINKED ISSUE #1741: Unit tests fail on non-english Locales because localized coordinate strings are expected
**Describe the bug**
The unit tests of mars_sim.core fail because the LunarColonyManager class can't be initialized. The root cause is here in Coordinates:

```
if (directionStr.compareToIgnoreCase(WEST_SHORT) == 0)
	longValue = 360D - longValue;
else if (directionStr.compareToIgnoreCase(EAST_SHORT) != 0)
	throw new IllegalStateException("Invalid Longitude direction : " + directionStr);
```

The constants are defined to be loaded from a resource using I13N:

```
private static final transient String SOUTH_SHORT = Msg.getString("direction.southShort");
private static final transient String WEST_SHORT = Msg.getString("direction.westShort");
```

Because of that, a hard coded coordinate in the Unit tests throws an exception, failing the creation of the LunarColonyManager and a whole chain of unit tests.

The unit tests pass if a correct coordinate is entered. The constants used for unit tests should NOT be depending on a specific locale being selected for running the unit tests, especially if hard-coded data or configuration file data is parsed.

**To Reproduce (Please provide steps)**

1. Switch to German locale, so that "direction.eastShort" gets defined as "O".
2. Run tests.

bevans2000:
This PR makes Coordinates parsing more independent of the language binding. It achieves this in a number of ways:

- Isolate output and parsing of Coordinates to new CoordinatesFormat class
- Create a new Decimal style of Coordinate that avoid use of compass names
- Decimal style is used in scenario templates
- Reduce the use of the Coordinate constructor that takes 2 Strings

close #1741

### DIFF
--- a/mars-sim-core/src/main/java/com/mars_sim/core/map/location/CoordinatesFormat.java
+++ b/mars-sim-core/src/main/java/com/mars_sim/core/map/location/CoordinatesFormat.java
@@ -62,6 +62,31 @@ public final class CoordinatesFormat {
     }
 
     /**
+     * Parse latitude string to phi, throwing an unchecked exception if it fails.
+     * @param latitude
+     * @return
+     */
+    static double parseLatitude2PhiUncheck(String latitude) {
+        try {
+            return parseLatitude2Phi(latitude);
+        } catch (CoordinatesException e) {
+            throw new IllegalArgumentException(e.getMessage(), e);
+        }
+    }
+
+    /**
+     * Parse longitude string to theta, throwing an unchecked exception if it fails.
+     * @param longitude
+     * @return
+     */
+    static double parseLongitude2ThetaUncheck(String longitude) {
+        try {
+            return parseLongitude2Theta(longitude);
+        } catch (CoordinatesException e) {
+            throw new IllegalArgumentException(e.getMessage(), e);
+        }
+    }
+    /**
 	 * Parses a latitude string into a phi value. e.g. input: "25.344 N"
 	 * For latitude string: North is positive (+); South is negative (-)
 	 * For phi : it starts from the north pole (phi = 0) to the south pole (phi = PI)
--- a/mars-sim-core/src/main/java/com/mars_sim/core/map/location/CoordinatesFormat.java
+++ b/mars-sim-core/src/main/java/com/mars_sim/core/map/location/CoordinatesFormat.java
@@ -1,0 +1,235 @@
+/*
+ * Mars Simulation Project
+ * CoordinatesFormat.java
+ * @date 2025-09-16
+ * @author Barry Evans
+ */
+package com.mars_sim.core.map.location;
+
+import java.text.DecimalFormat;
+import java.text.ParseException;
+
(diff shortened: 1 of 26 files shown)


mars-sim-core/src/main/resources/messages.properties
```diff
@@ -286,15 +286,13 @@ ConstructionMissionCustomInfoPanel.column.material=Material
...
-Coordinates.error.latitudeBeginWith=Latitude must be between 0.00 and 90.00 (Note: both comma and period are valid decimal mark)
-Coordinates.error.latitudeEndWith=Latitude must end with the direction notation of ''{0}'' or ''{1}''
-Coordinates.error.latitudeMissing=Latitude is missing or not valid
-Coordinates.error.latitudeBadFormat=Valid latitude format is a two-decimal place number with a space and either N or S. e.g. '0.10 N' or '89,20 S'
-Coordinates.error.longitudeBeginWith=Longitude must be between 0.00 and 360.00 (Note: both comma and period are valid decimal mark)
-Coordinates.error.longitudeEndWith=Longitude must end with direction ''{0}'' or ''{1}''
-Coordinates.error.longitudeMissing=Longitude is missing or not valid
-Coordinates.error.longitudeBadFormat=Valid longitude format is a two-decimal place number with a space with either E or W. e.g. '0.10 E' or '179,20 W'
+
+CoordinatesFormat.error.latitudeRange=Latitude must be between 0.00 and 90.00 (Note: both comma and period are valid decimal mark)
```
mokun, line 290: Should we make the lat and lon in decimal format to stay at only 2 decimal places ?

Or in 4 decimal places ?

bevans2000:
@mokun I looked at ```ColonyManager``` and it has a lot of hardcoded Coordinate values. Looks like this should be in a config file.

mars-sim-core/src/main/java/com/mars_sim/core/map/location/CoordinatesFormat.java
```diff
@@ -0,0 +1,235 @@
...
+
+    private static final String DEG_SIGN = Msg.getString("direction.degreeSign");
+
+    private static final String NORTH = Msg.getString("direction.northShort");
+	private static final String EAST = Msg.getString("direction.eastShort");
+	private static final String SOUTH = Msg.getString("direction.southShort");
+	private static final String WEST = Msg.getString("direction.westShort");
+
+	/** Currently, lat and lon are up to 4 decimal places. Thus the decimal format '0.0000' is used. */
+	static final DecimalFormat DIGIT_FORMAT = new DecimalFormat("0.0000"); //$NON-NLS-1$
```
mokun, line 31: Is this for the degree format as well ?

bevans2000: Yes, this controls the number part only

Issue #356, opened 2021-06-06
Replace the console driven main menu with swing UI-based main menu


mokun:
We have been relying upon the [console](https://github.com/mars-sim/mars-sim/wiki/Mars-Net-Console) terminal for configuring the mars-sim before starting the simulation.

No doubt, in Headless Edition, we will continue to make use of the console as the major method for interaction. 

In any cases, setting up [Remote Connection]( https://github.com/mars-sim/mars-sim/wiki/Remote-Console-Connection) via SSH client is available as an option, which will pave the way for future networking capability of mars-sim.

Seems it's about time we rebuild the Main Menu using Swing UI for the Swing Edition to make it feel more complete.

Q: are there any packages (that have maven artifacts) out there that will help facilitate our job ?

Also, we should add back displaying a splash screen during the startup.

> Note : the caveat is that the console does use swing components to build because the [beryx / text-io](https://github.com/beryx/text-io) package that we depend on is using [JLine](https://github.com/jline/jline2).

bevans2000:
I like the terminal interface, gives a quick way to examine and change things. I will have a look when I get back home

bevans2000:
What do you mean by this? You want to see a startup menu that allows the user to select what to start? The command line args do drive it but I guess you are talking about the novice user

mokun:
If the player doesn't bypass the startup, it will go into the Main Menu in the console window.

```
 ---------------  M A R S   S I M U L A T I O N   P R O J E C T  ---------------

                          * * *   Main Menu   * * *


0. Exit
1. Start a new Sim
2. Load a saved Sim
3. Change screen resolution

Choose an option : ​
```

This main menu is in text and not made in swing UI for players.

mokun:
See the logs : 

```
2025-09-20T16:00:03.5 (Config) MarsProject : Starting Mars Simulation Project pre-3.10.0 - 64-bit Java 21.0.7 - 10 Cores
2025-09-20T16:00:03.5 (Info) SimulationConfig : Loading simulation configuration from /Users/spacebear/.mars-sim/xml
2025-09-20T16:00:03.9 (Config) MarsProject : Do you want to bypass the console menu and start a new default simulation in Sandbox Mode ?
2025-09-20T16:00:03.9 (Config) MarsProject : To proceed, please choose 'Yes' or 'No' button in the dialog.
2025-09-20T16:00:06.4 (Config) MarsProject : Please go to the Console Main Menu to choose an option.
[...]

PR #1759, opened 2025-09-28
Clean up terminal code and move to UI module

bevans2000:
Changes:
- Move Swing based terminal classes into the UI module
- Combine InteractiveTerm into MarsTerminal
- UIConfig uses the same structure for main window & terminal
- MarsTerminal handles positioning itself
- Increase width of update panel in splash screen
- Move textIO module from console to UI module
- 
- Ref: #356


mokun:
So if someone wants to develop UI with LibGDX or JavaFX, will they still be able to use the beryl console ?

mokun:
> But what does the Commander Profile class do? 

It's for the so called *Command Mode*, meaning the player is seated as the commander of a single settlement scenario. There are AI settlements out there but he can't manage them.

bevans2000:
> Does it mean no more mars-sim-console eventually ?

No, the console module is about a conversation with the user. The delivery of that conversation takes different formats and may change it the future

- Mars terminal is for a Swing based conversation 
- Command line is used for the Headless version
- A remote SSH conversation can be used

But the commands and interaction logic is what stays in the Console module.

mokun:
> Yes I get that but how is it used? It's never called by anything

Have you used the command mode before ? 

The command mode did work for a while. 

I used it many times to play-test simulations initially since it's the most *distraction free* environment. I only have to focus one the development of one single settlement, while other AI settlements will exist.

But afterward, we haven't followed up on updating it to make it compatible continuously.

Issue #1624, opened 2025-07-24
Revamp the area collision algorithm


mokun:
**Describe the bug**
- Area collision methods uses up 10 to 40% CPU in a typical simulation
- Currently, there are 3 major applications of calculating area collision :  
1. Settlers walking outside
2. Vehicles finding parking space
3. Building placement

**Expected solution**
- Use a more efficient search or pathfinding algorithm 
- Simplify methods in `LocalAreaUtil` and `WalkOutside`

**Java classes involved**
- `WalkOutside`, `LocalAreaUtil`

- In `WalkOutside`, we have 
```java
	
	/**
	 * Determines the outside walking path, avoiding obstacles as necessary.
	 *
	 * @return walking path as list of X,Y locations.
	 */
	private List<LocalPosition> determineWalkingPath() {

		List<LocalPosition> result = new ArrayList<>();
		result.add(start);

		// Check if direct walking path to destination is free of obstacles.
		Line2D line = new Line2D.Double(start.getX(), start.getY(), destination.getX(), destination.getY());

		boolean freePath = LocalAreaUtil.isLinePathCollisionFree(line, worker.getCoordinates(), true);

		if (freePath) {
			result.add(destination);
		}

		else {
			// Determine path around obstacles using A* path planning algorithm.
			List<LocalPosition> obstacleAvoidancePath = determineObstacleAvoidancePath();
		
			if (obstacleAvoidancePath != null) {
				// Set to obstacle avoidance path.
				result = obstacleAvoidancePath;
			}

			else {
				// Accept obstacle-blocked path as last resort.
				result.add(destination);
				obstaclesInPath = true;
			}
		}

		return result;
	}

	/**
	 * Determines obstacle avoidance path.
	 *
	 * @return path as list of points or null if no path found.
	 */
	private List<LocalPosition> determineObstacleAvoidancePath() {

		// Check if start or destination locations are within obstacles.
		// Return null if either are within obstacles.
		boolean startLocWithinObstacle = !LocalAreaUtil.isPositionCollisionFree(start, worker.getCoordinates());
		if (startLocWithinObstacle) {
			//logger.warning(worker, "Start/End positions are inside an entity");
			return null;
		}
		
		boolean destinationLocWithinObstacle = !LocalAreaUtil.isPositionCollisionFree(destination, worker.getCoordinates());
		if (destinationLocWithinObstacle) {
			//logger.warning(worker, "Start/End positions are inside an entity");
			return null;
		}
	
[...]

PR #1768, opened 2025-10-05
Extract outside path finder logic

bevans2000:
This is a PR to prepare the way to change the path finder algorithm in the future. It isolates the logic into a separate class.

Changes:
- Extract path finder logic from WalkOutside behind a new interface OutsidePathFinder
- Existing logic move to CollisionPathFinder as it is based on collision detection
- Path logic held in a new dedicated walk package
- Add new method to LocalAreaUtil to hide the use of Line2D
- Create new unit test for OutsidePathFinder

Ref #1624


bevans2000:
Proved harder than expected. Tests were failing on GitHub but working locally.

Issue #1654, opened 2025-08-08
Show Historical Events and Missions for one settlement/sponsor


mokun:
**Is your improvement request related to a situation? **
- At sol 100+, both Historical Events and Missions tab in Monitor Tool can be an overwhelmingly long list of events/missions.
- There is no Authority Combobox on these two tabs for players to choose one settlement/sponsor.

**Describe goal(s) you have in mind**
- Add an Authority Combobox on top center for both the historical Events and Missions tab.
- Add checkboxes at the bottom such as Medical, Hazard, Malfunctions, etc. to allow players focus on one or more categories at a time.

**Specifications**
- pre-3.10.0 

**Related Issues and Discussions**
- [Settlement Combobox](https://github.com/mars-sim/mars-sim/issues/524)

**Additional context**
- Previously, it has been suggested we can potentially add an "All" option to the Authority/Settlement Combobox for Citizens/Robots/Traded Good/Food/Vehicles tab.

bevans2000:
It's not a difficult change but I do this the historical event logic need revisiting to make to more easy to use in other places

wonderinglostsoul44:
Add category filters in “Historical Events & Missions”

Why: Long lists need quick filtering by authority and event category; the issue calls for an Authority combo and checkboxes. Start with a tiny category enum + filter hook.

What to change (model only; UI can follow):

// In your event model file (keep it small)
public enum EventCategory { MEDICAL, HAZARD, MALFUNCTION, MISSION, LOGISTICS }

// In the table model / list model for the Monitor Tool
private EnumSet<EventCategory> filter = EnumSet.allOf(EventCategory.class);

public void setCategoryFilter(EnumSet<EventCategory> f) {
    this.filter = EnumSet.copyOf(f);
    refilterAndRefresh();
}

private void refilterAndRefresh() {
    this.visibleEvents = this.allEvents.stream()
        .filter(ev -> filter.contains(ev.category()))
        .collect(Collectors.toList());
    fireTableDataChanged();
}


Why this helps:
You’ll be ready to wire the Authority combobox and category checkboxes later, but users get immediate value by reducing clutter now.

mokun:
> Yes I agree the events are a bit of a problem. They need a better and more flexible structure.

Well we could.

> The Event type enum is a bottleneck and doesn't allow us to easily add events. 

@wonderinglostsoul44 

We may not want to use EnumSet since it will limit us down the road.

[...]

PR #1781, opened 2025-10-19
Historical Events refactor

bevans2000:
This pr addresses should changes to the historical event logic to simplify and improve the use.

- `HistoricalEvent` holds reference to the Coordinate and Settlement instead of the string version. It allows better filtering
- `EventType` is renamed and moved to `HistoricalEventType` to be consistent with other classes for events
- Each event type had the related parent category embedded instead of separate.
- The subclasses of the historical event are removed, all events use the same class.
- Arrival of anew settlement creates the event against the new settlement.

This a 3.11 piece of work and not for 3.10.

Follow on from #1654


mars-sim-core/src/main/java/com/mars_sim/core/events/HistoricalEventManager.java
```diff
@@ -32,12 +32,13 @@ public class HistoricalEventManager implements Serializable {
 	 * This defines the maximum number of events that are stored. It should be a
 	 * standard property.
 	 */
-	private final static int TRANSIENT_EVENTS = 50;
+	private static final int TRANSIENT_EVENTS = 50;
```
mokun, line 35: Nor that we can easily go to sol 200, we could scale that to looking back at the last 200 events.

mars-sim-core/src/main/java/com/mars_sim/core/interplanetary/transport/Transportable.java
```diff
@@ -134,8 +135,9 @@ public void setArrivalDate(MarsTime arrivalDate) {
 
 	/**
 	 * Performs the arrival of the transportable.
+	 * @return Return an event that describes the arrival
 	 */
-	protected abstract void performArrival(SimulationConfig sc, Simulation sim);
+	protected abstract HistoricalEvent performArrival(SimulationConfig sc, Simulation sim);
```
mokun, line 140: Q: is the arrival of a transport event the same as the arrival of the resupply event ?

mars-sim-core/src/main/java/com/mars_sim/core/interplanetary/transport/resupply/Resupply.java
```diff
@@ -157,7 +160,7 @@ public ResupplySchedule getTemplate() {
 	 * Generates the delivery event.
 	 */
 	@Override
-	public synchronized void performArrival(SimulationConfig sc, Simulation sim) {
+	public synchronized HistoricalEvent performArrival(SimulationConfig sc, Simulation sim) {
```
mokun, line 163: May be we can deconflict the setting up of the resupply mission as a 'launch' event, while the transport manager handles the 'arrival' or landing event, or the other way around ? !

mars-sim-core/src/main/java/com/mars_sim/core/person/ai/mission/AbstractMission.java
```diff
@@ -378,7 +379,7 @@ private final void memberLeave(Worker member) {
 	      		person.getShiftSlot().setOnCall(false);
 	      	}	
 
-			registerHistoricalEvent(person, EventType.MISSION_FINISH, "Removing a member");
+			registerHistoricalEvent(person, HistoricalEventType.MISSION_FINISH, "Removing a member");
```
mokun, line 382: Guess it's good to record the successful completion of a mission. 

But will removing a member as an event needed ?

mars-sim-core/src/main/java/com/mars_sim/core/person/ai/mission/RoverMission.java
```diff
@@ -434,7 +433,7 @@ private boolean checkMembership(Worker member, Rover r) {
 			}
 			
 			MissionStatus status = MissionStatus.createResourceStatus(MISSION_LEAD_NO_SHOW.getName());
-			abortMission(status, EventType.MISSION_LEAD_NO_SHOW);
+			abortMission(status, HistoricalEventType.MISSION_LEAD_NO_SHOW);
```
mokun, line 436: This should happen from time to time but I don't recall seeing it. May be it needs fixing.

Having said that, we'll need the ability to natively filter the type of historical events to show in the Event tab in Monitor Tool.

mokun (approved):

PR #1791, opened 2025-10-21
Correct numerous UI problems

### LINKED ISSUE #1776: Fix naming an Exploration's navpoint. Give each an interesting name to remember
**Describe the bug**
- Recently we have re-enabled the ability to [collect rocks](https://github.com/mars-sim/mars-sim/commit/aac02e965e8cdd2a333c24dab43ed20c20574982) in Exploration Mission, it has been noticed that there are sites named **Unknown** that are being populated in an Exploration mission from time to time (but not all the time). Need to pin down this bug that may have been introduced fairly recently. 

*Associated Improvement**
- Apply creativity in naming a site. Drop the boring convention of naming site as "Exploration Site 1", "Exploration Site 2", etc. 
  - May adopt a similar playbook set forth by NASA in their surface rover missions in naming rocks as "Scooby Doo", "Barnacle Bill", etc. 
- Make each Exploration's navpoint *permanent* (if not done so) in a simulation. 
   - It makes more sense for every serious explorer come back to each of these newly explored navpoints for further in-depth investigations. Only tourists want an one-off experience.

**Java classes involved**
- `Exploration`, `NavPoint`

In `AbstractVehicleMission`,
```java
	/**
	 * Gets the current navpoint the mission is stopped at.
	 * 
	 * @return the description of the current navpoint.
	 */
	public final String getCurrentNavpointDescription() {
		if (travelStatus != null && AT_NAVPOINT.equals(travelStatus)
			&& navIndex < navPoints.size()) {
			return navPoints.get(navIndex).getDescription();
		}
		
		return "Unknown";
	}
```

**Screenshots**

- Had two sites but the first one was not named.
Image

- Only one site but it was not named properly.
Image

**Specifications**
- pre-3.10.0

bevans2000:
This fixes 3 problems

1. Exploration sites have wrong names
2. Monitor Tool filter can raise NPE
3. DateDialog not appearing on OrbitViewer

Ref: #1778 
close #1776

### DIFF
--- a/mars-sim-fxgl/src/main/java/com/mars_sim/fxgl/InventorySample.java
+++ b/mars-sim-fxgl/src/main/java/com/mars_sim/fxgl/InventorySample.java
@@ -10,6 +10,7 @@ import com.almasb.fxgl.app.GameSettings;
 import com.almasb.fxgl.entity.Entity;
 import com.almasb.fxgl.input.UserAction;
 import com.almasb.fxgl.inventory.Inventory;
+import com.almasb.fxgl.inventory.view.InventoryView;
 import com.almasb.fxgl.scene.SubScene;
 import javafx.scene.control.Button;
 import javafx.scene.input.KeyCode;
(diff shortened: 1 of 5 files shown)


mokun:
> changes to the JavaFX version breaks the build. I've rolled them back.

Thanks ! I switched to the new FXGL version to see how well it works. But it's incompatible with the existing codes. Forgot to change it back.

PR #1793, opened 2025-10-25
Add build timestamp to Version

bevans2000:
Changes:
- Version now has a build timestamp field
- core module pom updated so plugin creates timestamp
- MainWindowMenu shows build timestamp in version dialog
- Remove the incorrectly placed constants from MarsTermninal
- Remove excessive log from MasterClock


mokun:
That menu item within mars-sim for comparing the latest mars-sim that is available for download and the version of mars-sim that a player is running. 

Yesterday I added an url for plate to manually check for the latest version.

In future the having should be accomplished at the click of a button.

bevans2000:
I'm away now so won't be making any changes

Issue #1730, opened 2025-09-06
Create a Governance Model


bevans2000:
## Problem
The population size of a Settlement influences the Roles needed in a Settlement. And consequently the approval/sign off for various activities, e.g. Mission Review, Job Changes.
This feature is used in numerous areas within the Simulation, UI Organization Panel, Chain of Command, Review Mission Plan etc.
However each component currently re-implements the a subset of the rules leading to duplicate code and potential for different components to have different inconsistent rules.

## Solution
The solution is to centralize the rules into the existing ```ChainOfCommand``` class and more specifically standardize the rules into a new configuration class called ```GovernanceModel```. This new class will encapsulate the various rules and be a stateless configuration class. Settlements would select the appropriate Governance model based on the population size by default.
Furthermore, the various Governance models could be configured externally.
The benefits of this change will allow a more controlled management of Settlement Governance and in the future could lead to shared Governance between all Settlements in the same Reporting Authority.

mokun:
> GovernanceModel. This new class will encapsulate the various rules and be a stateless configuration class.

Sure. Let's try that.

Is there something equivalent to this that we already have ? 

Is the EntityModel being one of it ?

> in the future could lead to shared Governance between all Settlements in the same Reporting Authority

This sounds fabulous. 

Say there's a medical emergency in Rover 1, Settler X was ill, but Rover 1 is nowhere near its home, which is Settlement A.

However, Rover 1 is near Settlement B.

The GovernanceModel would come in handy. Say if both settlements have an prior agreement in sharing resources, Settlement B would dispatch a passenger drone to Rover 1 to pick up Settler X and take him back to Settlement B for treatment.

mokun:
> it can be used to decide which job types should do scientific studies which is controlled by initial population.

Can we make it *configurable* by player ? 

We can use a combo box for each configurable authority/permission. e.g. What role/job should authorize scientific studies, who should allow construction mission, what role have certain types of settlement jurisdiction, etc.

This will richly diversify the way a settlement operates under each sponsor.

Some sponsors such as those in East Asia may require more central form of command structure, compared to the EU or US, etc.

I imagine each sponsor has its own preference as well.

We have a command structure tab in SettlementUnitWindow but it only shows a tree of hierarchy. It's a waste.

May be we can put it into some use somehow.

Urwumpe:
How do we resolve it, if rules contradict each other?

mokun:
[...]

PR #1799, opened 2025-11-09
Create Governance model

bevans2000:
This PR implements a centralised Governance model. It is a combination of clean up/centralisation of existing code with new features.

This delivers:

- Separate `GovernameRules` model
- Logic to govern Settlement is now defined by GovernanceRules class and not spread around code
- `ChainOfCommand` allocates all Council Roles, not just the top 2. This is important for larger Settlements.
- Organisation tab in Settlement is totally data driven from Governance Rules

Ref #1730

### DIFF
--- a/mars-sim-core/src/main/java/com/mars_sim/core/person/ai/role/RoleType.java
+++ b/mars-sim-core/src/main/java/com/mars_sim/core/person/ai/role/RoleType.java
@@ -62,6 +62,7 @@ public enum RoleType implements Named {
 		return this.level;
 	}
 
+	@Override
 	public final String getName() {
 		return this.name;
 	}
--- a/mars-sim-core/src/main/java/com/mars_sim/core/structure/ChainOfCommand.java
+++ b/mars-sim-core/src/main/java/com/mars_sim/core/structure/ChainOfCommand.java
@@ -42,7 +42,9 @@ public class ChainOfCommand implements Serializable {
 	private static final Logger logger = Logger.getLogger(ChainOfCommand.class.getName());
 
 	/** A record to store the maximum and current number of a role. */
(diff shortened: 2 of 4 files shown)


mars-sim-core/src/main/java/com/mars_sim/core/person/ai/role/RoleType.java
```diff
@@ -93,4 +81,107 @@ public boolean isVisitor() {
...
+	    };
+	}
+
+	/**
+	 * Returns the list of required skills for the given role type.
+	 * If the role type is not a Chief, null is returned.
+	 * @param roleType
+	 * @return
+	 */
+	public static List<SkillType> getRequiredSkills(RoleType roleType) {
```
mokun, line 114: Q: should a specialist of that division also have those designated skill types ?

mars-sim-core/src/main/java/com/mars_sim/core/person/ai/task/MeetTogether.java
```diff
@@ -287,18 +269,17 @@ private void clearTask() {
      * @param roleType
      * @return a set of persons
      */
-    public Set<Person> getPool(Collection<Person> ppl, RoleType roleType) {
-    	RoleType candidateType = RoleUtil.getChiefSpeciality(roleType);
+    private Set<Person> getPool(Collection<Person> ppl, RoleType roleType) {
+    	RoleType candidateType = RoleType.getChiefSpeciality(roleType);
  
-        Set<Person> pool = new UnitSet<>();
```
mokun, line 293: Q: how would the return of this method behave if, say, the serialization happens right at the moment this method is being executed ? 
 
When deserializing from a saved sim, will it return a set with valid person's object reference ?

PR #1810, opened 2025-11-30
Create EntityContentPanel for EquipmentUnitWindow

### LINKED ISSUE #1801: Create dedicated entity window for Reporting Authority
The Authority tab in the Person and Settlement windows contains a lot of information that is compressed into a single tab. There are numerous issues that will lead to more importance on the Authority.

The change here is to separate the presentation of Reporting Authority into a dedicated entity window. This will also be aligned with the standardization of entities having a dedicated details window.

### LINKED ISSUE #1807: Create EntityContentPanel
The current Swing framework links the unit window directly to the windowing framework. To support different windowing solutions like docking and JavaFX this needs breaking.
Also many of the individual tab panels have logic the should be centralised, e.g unit listeners.

**Describe the solution you'd  to see ##
Create a new `EnitityContentPanel` that subclasses the generic `ContentPanel` that renders a single `Entity` including configuring listeners. 
This will support forwarding any Entity events to the tab panels as well as Clock pulse updates.
This will form the base for all viewing on single Entities.

This will also need a change to the `MainDesktopPane` as the created panels will hosted in the generic ContentWindow

bevans2000:
This PR creates new EntityContentPanel but it still supports the existing UnitWindow pattern. It starts a transition phase supporting both old and new patterns 

Key Changes:
- Create EntityContentPanel class subclassing ContentPanel.
- EntityContentPanel supports adding EntityTabPanels mimicking UnitWindow logic.
- MainDesktopPane can support EntityContentPanel for the showDetails method but it has to support old and new style of windows.

EquipmentUnitWindow to used as the test case. Some common TabPanels are converted to EntityTabPanels and hence use UIContext instead of MainDesktopPane.

In addition a new Authority window is created to show details about an `Authority`

Close #1807
Close #1801

### DIFF
--- a/mars-sim-ui/src/main/java/com/mars_sim/ui/swing/StyleManager.java
+++ b/mars-sim-ui/src/main/java/com/mars_sim/ui/swing/StyleManager.java
@@ -597,6 +597,14 @@ public class StyleManager {
     }
 
     /**
+     * Gets the sub title font.
+     * @return
+     */
+    public static Font getSubTitleFont() {
+        return subTitleFont;
+    }
+
+    /**
      * Creates a standardized empty border.
      */
     public static Border newEmptyBorder() {
--- a/mars-sim-ui/src/main/java/com/mars_sim/ui/swing/entitywindow/authority/TabPanelGeneral.java
+++ b/mars-sim-ui/src/main/java/com/mars_sim/ui/swing/entitywindow/authority/TabPanelGeneral.java
@@ -20,7 +20,6 @@ import javax.swing.border.EmptyBorder;
 import com.mars_sim.core.authority.Authority;
 import com.mars_sim.core.tool.Msg;
 import com.mars_sim.ui.swing.ImageLoader;
-import com.mars_sim.ui.swing.StyleManager;
 import com.mars_sim.ui.swing.UIContext;
 import com.mars_sim.ui.swing.entitywindow.EntityTabPanel;
 import com.mars_sim.ui.swing.tool.svg.SVGIcon;
@@ -90,7 +89,7 @@ class TabPanelGeneral extends EntityTabPanel<Authority> {
 				
 		// Base details
 		var attrPanel = new AttributePanel();
-		attrPanel.setBorder(StyleManager.createLabelBorder("Details"));
+		attrPanel.setBorder(SwingHelper.createLabelBorder("Details"));
 		mainPanel.add(attrPanel);
 
 		attrPanel.addTextField("Code", agencyShortName, null);
--- a/mars-sim-ui/src/main/java/com/mars_sim/ui/swing/unit_window/TabPanel.java
+++ b/mars-sim-ui/src/main/java/com/mars_sim/ui/swing/unit_window/TabPanel.java
@@ -24,6 +24,7 @@ import com.mars_sim.core.Simulation;
 import com.mars_sim.core.Unit;
 import com.mars_sim.ui.swing.MainDesktopPane;
 import com.mars_sim.ui.swing.StyleManager;
+import com.mars_sim.ui.swing.utils.SwingHelper;
 
 @SuppressWarnings("serial")
 public abstract class TabPanel extends JScrollPane {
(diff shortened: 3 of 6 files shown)


mars-sim-ui/src/main/java/com/mars_sim/ui/swing/unit_window/InventoryTabPanel.java
```diff
@@ -411,14 +352,18 @@ public Class<?> getColumnClass(int columnIndex) {
...
 
+		@Override
         public String getColumnName(int columnIndex) {
 			// Internationalized and capitalized column headers
-            if (columnIndex == 0) return Msg.getString("InventoryTabPanel.Resource.header.name");
-            else if (columnIndex == 1) return Msg.getString("InventoryTabPanel.Resource.header.quantity");
-            else if (columnIndex == 2) return Msg.getString("InventoryTabPanel.Resource.header.capacity");
-            else return "unknown";
+            return switch (columnIndex) {
+              case 0 -> Msg.getString("InventoryTabPanel.Resource.header.name");
```
mokun, line 357: I hope getColumnName() doesn't get called all the time since Msg.getString("InventoryTabPanel.Resource.header.name") can be a private string and in doing so, there's no need of constantly having to look up message.properties, you know.

mars-sim-ui/src/main/java/com/mars_sim/ui/swing/unit_window/InventoryTabPanel.java
```diff
@@ -541,26 +487,28 @@ public Class<?> getColumnClass(int columnIndex) {
...
+		@Override
         public String getColumnName(int columnIndex) {
 			// Internationalized and capitalized column headers
-            if (columnIndex == 0) return Msg.getString("InventoryTabPanel.item.header.name");
-            else if (columnIndex == 1) return Msg.getString("InventoryTabPanel.item.header.quantity");
-            else if (columnIndex == 2) return Msg.getString("InventoryTabPanel.item.header.mass");           
-            else if (columnIndex == 3) return Msg.getString("InventoryTabPanel.item.header.reliability");
-            else return Msg.getString("InventoryTabPanel.item.header.mtbf");
+            return switch (columnIndex) {
+              case 0 -> Msg.getString("InventoryTabPanel.item.header.name");
```
mokun, line 492: Too much key value lookup

mars-sim-ui/src/main/java/com/mars_sim/ui/swing/unit_window/MaintenanceTabPanel.java
```diff
@@ -294,7 +305,7 @@ public String getColumnName(int columnIndex) {
 			case 5:
 				return Msg.getString("MaintenanceTabPanel.header.probability"); //$NON-NLS-1$
```
mokun, line 306: You can eliminate the key value lookup if we don't want to

mars-sim-ui/src/main/java/com/mars_sim/ui/swing/entitywindow/authority/AuthorityWindow.java
```diff
@@ -0,0 +1,30 @@
...
+
+import com.mars_sim.core.authority.Authority;
+import com.mars_sim.ui.swing.UIContext;
+import com.mars_sim.ui.swing.entitywindow.EntityContentPanel;
+
+/**
+ * The AuthorityWindow is the window for an authority entity.
+ * It contains a number of tabs.
+ */
+public class AuthorityWindow extends EntityContentPanel<Authority> {
```
mokun, line 19: Great that we have AuthorityWindow now.

bevans2000: It can be access via the Organisation Tab in the Settlement Window or the General Tab from Person.
I need to add a link from the ArrivingSettlement panel as well.

mars-sim-ui/src/main/java/com/mars_sim/ui/swing/unit_window/InventoryTabPanel.java
bevans2000: The column names are only used to initial renderer the header. But it's a good point.

PR #1829, opened 2025-12-13
Fix NPE in Colonist

### LINKED ISSUE #1828: NPE in the Colony code
**Describe the bug**
There is a NPE thrown in the Colony code when research projects are in play.

When the number of research projects change in a `Colony` a `Colonist` is removed from the `Population` and stored against the `Nation` pool.
The research project increase and the Colonist is reassigned back to the Colony but the Colonist is not assigned to a Colony. This will then generate a NPE on the next clock pulse.
The code clearly is wrong but it is unclear why this problem has not been uncovered previously.

**Discuss any potential problems you have and have not ruled out**
The solution is the `setColony` method must be applied when retrieving a Colonist from a Nation.
However the Colony logic does currently not impact the user's experience, it generates log but no visual representation nor does it have any indrect impact.
Therefore a flag will also be added to the Scenario concept that defines how many Colony are to be created. By default this will be zero.

bevans2000:
This PR has 2 purposes:
- Fix the NPE in Colonist after it has been reassigned to a Colony
- Allow the number of Colonies created to be controlled by the Scenario

Close #1828

### DIFF
--- a/mars-sim-core/src/main/java/com/mars_sim/core/moon/LunarColonyManager.java
+++ b/mars-sim-core/src/main/java/com/mars_sim/core/moon/LunarColonyManager.java
@@ -107,7 +107,7 @@ public class LunarColonyManager implements Serializable, Temporal {
 	public boolean timePassing(ClockPulse pulse) {
 		lunarWorld.timePassing(pulse);
 		
-		if ((colonies.size() < maxColonies) && (RandomUtil.getRandomInt(100) <= 5)) {
+		if ((colonies.size() < maxColonies) && (RandomUtil.getRandomInt(100) == 1)) {
 			addColony();
 		}
 					


mokun:
One initial goal was to create a pool of researchers in moon colonies (and on various nations on earth) who could collaborate on scientific studies with those on Mars.

It can potentially bring in 

Turns out it was hard to implement.

What do you see we can do to make that happen eventually ?

This is quite needed for simulation scenarios with just one settlement on Mars with 4 or 8 settlers. Scientific studies couldn't get started when the Mars population is so low.

But since then, I have seen passes just this potential benefit. 

For instance, we can begin to simulate the economic strength of each participating nation on earth and could in future simulate some aspects of the entire supply chain for making parts and equipment and launching colonization or resupply missions between earth, moon and Mars.

bevans2000:
I think first decision is how does this impact the simulation. Then we have to give it a presence in the UI. The big reworking of the Swing UI I'm doing will be perfect to support showing Colonies in the Monitor tool and then a dedicated window.
I don't think we necessarily need to model the individual Colonists, maybe we could just model the number of them

PR #1832, opened 2025-12-14
Remove the Nation holding Colonists

### LINKED ISSUE #1828: NPE in the Colony code

bevans2000:
The Population class has a potential to generate random ClassCastException as it pulls Colonists from the Nation to be reused.

This removes the pooling of Colonists in the Nation class.

Close #1828


mars-sim-core/src/main/java/com/mars_sim/core/authority/Nation.java
```diff
@@ -8,99 +8,26 @@
...
-	/**
-	 * Gross Domestic Product (GDP) is the standard measure of the value 
-	 * added created through the production of goods and services 
-	 * in a country during a certain period. 
-	 * 
-	 * As such, it also measures the income earned from that production, 
-	 * or the total amount spent on final goods and services 
-	 * (less imports).
-	 */
-	private double GDP;
```
mokun, line 44: Can we have the GDP, PPP, annualGDPGrowth, GDPPerCapita, etc. back ?

mokun: We will be loading these values from the xls/json file of that nation. They are considered useful params showcasing the economical capabilities of a nation.

bevans2000: > Can we have the GDP, PPP, annualGDPGrowth, GDPPerCapita, etc. back ?

None of the values were used and just generated code smells. Add them back in once there is a use for them I think

Issue #1853, opened 2026-01-17
Consolidate the rendering of numbers in the Tab Panels


bevans2000:
## Problem

Many of the tab panels display `double`  values in JLabels. Each rendering involves holding a cached value and using a DecimalFormat to convert the double into a text value.
Whenever the source value is potentially updated, the new value is compared to the cache and if difference reformatted into the JLabel. The same problem applies to the rendering of integers as well.

This pattern generates a lot of duplicate code within the tab panels and reduce the readability of the code.

## Proposal

The proposal is to create a specialisation of the JLabel that works at the `double` or `integer` level and handles caching and styling within the class instead of the parent tab panel. This simplifies the code within the tab panels, increasing readability.
Applying the pattern is repetitive and hence a perfect case for using CoPilot.

PR #1854, opened 2026-01-17
feat: Simplify rendering of doubles in Tab Panels

bevans2000:
This PR simplifies the rendering of double values in the `EntityTabPanel` class. It moves the caching, styling and comparision check into a new `JDoubleLabel` class. This simplifies the parent code and improves the readability.

This has been applied to all the relevant tab panels.

Ref: #1853

### DIFF
--- a/mars-sim-ui/src/main/java/com/mars_sim/ui/swing/components/JDoubleLabel.java
+++ b/mars-sim-ui/src/main/java/com/mars_sim/ui/swing/components/JDoubleLabel.java
@@ -1,6 +1,6 @@
 /*
  * Mars Simulation Project
- * JFormatLabel.java
+ * JDoubleLabel.java
  * @date 2026-01-11
  * @author Barry Evans
  */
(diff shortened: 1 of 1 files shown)


mars-sim-ui/src/main/java/com/mars_sim/ui/swing/astroarts/DateDialog.java
```diff
@@ -67,7 +67,7 @@ public DateDialog(OrbitViewer viewer, ATime atime, LocalDateTime earthTime) {
 
 		currentPanel.setBorder(StyleManager.createLabelBorder("Current Date"));	
 
-		AttributePanel attrPanel = new AttributePanel(3, 1);
+		AttributePanel attrPanel = new AttributePanel();
```
mokun, line 70: Q: Is there a way to make the left column that holds the text to be wider by default ?
 
The right column usually holds numbers and doesn't really need that much space and can be narrow to save space .

mokun (approved):

mars-sim-ui/src/main/java/com/mars_sim/ui/swing/astroarts/DateDialog.java
bevans2000: Yes it's a problem with the 50/50 approach. It's magnified in the Docking layout because the panels are generally wider and hence more space is wasted.
I am experimenting with a dynamic layout version that calculates the best number of columns based on the size of the panel and width of the labels and values. I will look to close it out and push it.

bevans2000: Yes it's a problem with the 50/50 approach. It's magnified in the Docking layout because the panels are generally wider and hence more space is wasted.
I am experimenting with a dynamic layout version that calculates the best number of columns based on the size of the panel and width of the labels and values. I will look to close it out and push it.

Issue #1859, opened 2026-01-24
Convert the MissionTool to follow the EntityContent pattern


bevans2000:
## Problem

The current MissionTool mixes the ability to browse Missions as well as the ability to view and create new Missions. The recent Swing rework project has laid down a strong user interaction framework based around many navigation links to leading to a single means to render Entities.

The larger "tool" based approach does not lead itself to the dockering window layout which favours numerous smaller/compact windows.
There is specific code in place to handle the navigation to the Mission tool which complicates the transition to being window agnostic.

## Proposal

The proposal is to replace the single MissionTool with 2 new windows:

- A Mission Window that follows the EntityContent approach to render all properties of a Mission. This would follow all the existing navigation links to Entity windows.
- A new dedicated Mission create window following a wizard style of approach.

The existing tree navigation of Missions in the MissionTool will be dropped as Missions will be able to be viewed through the numerous Entity navigation paths.

The create windows will create a generic wizard framework based on the ContentPanel approach. This can be applied to the SupplyWindow at a later dates.

bevans2000:
I mean the navigation panel on the create mission wizard when selecting a site for ice collection

bevans2000:
Yes that's odd, create a new issue for it to keep it separate. 
The rules should be in the Governance class but initially I would say only one active Study per Science type per Settlement

PR #1867, opened 2026-02-07
feat(ui):  Create standalone Mission Create wizard

bevans2000:
This isolates the Create Mission Wizard out of the Mission Tool for Missions that do not need a Map selection. This needs a number of high level changes:

- Create a WizardPane framework where the goal is centralise all common logic.
  - Handling of minimal data entry on each step
  - Next and Previous routing
  - Cancel or final Finish operation
- The wizard is based on steps where each step focus on a specific action
- The sequence of steps to be applied is defined within the wizard. 
- Wizard provides a specialised Step to select from a model of items.
  - All selection and control of the table is included
  - Step model supports failing cells that do not meet selection criteria with reason displayed a tooltips
  - The table updates as the underlying `MonitorableEntity` changes
 - Instructions are displayed in each step.

The wizard allows the existing Mission wizard to be converted new generic approach. This results in a smaller and simpler code base. It also allows panels to be shared between step sequences.

There are some changes to the core to support this:

- `MineralSite` now has a name attribute. This improves the user experience and monitoring.

Ref #1859


bevans2000:
Yes I want to allow a Mission to be tweaked but it will still be limited changes.

I'm determined to get the Docking style alternative working.

Issue #1808, opened 2025-11-29
Adopt docking window style based on content panel framework


bevans2000:
Deliver a Swing UI using a docking window style, i.e. IDE, following the discussion in #979.
This will look to use the https://github.com/andrewauclair/ModernDocking framework.

mokun:
Awesome !

> use the -dockingui command line argument it will work with the new approach.

Is there a way during the simulation to switch between the normal mode vs. the docking approach ?

bevans2000:
Potentially yes as the same Simulation object would be kept. But more likely control it via a Preference setting. 
Got to work out the LAF first and convert Settlement and Building Unit Window to the Content Panel approach.

mokun:
So when you say 'browser' what you mean is a 'main object tree panel', right?

bevans2000:
Yes, might need a better name.

bevans2000:
Yes for sure. This Paul Request is a work on progress. I wanted to get it out there so you can give it a try. There is still work but it's looking good

mokun:
How do I start it in docking window style ?

mokun:
Got it. using -dockingui.

One problem is that if the Event browser is open, playing the individual sound clip (not the music) of some sort is non-stop.

Why is it link to any sound ?

bevans2000:
It plays a sounds whenever an event is created in the Event Browser

PR #1901, opened 2026-04-06
feat(ui): Change Windowing style via settings menu

bevans2000:
Changes:
- Add docking ui property to UIConfig.
- Add radio buttons to settings menu for switching between window and docking UI.
- MarsProject uses the UIConfig to decide window style
- Command line can force docking UI with --use-docking-ui
- UIConfig supports both windowing styles

Ref #1808


bevans2000:
@mokun This is will now provide the user with the ability to run either the Docking style or normal Window style.
It's marks the end of the Swing rework project, just to sweep up any problems.

mokun (approved):

PR #1903, opened 2026-04-12
fix(ui): Docking layout

### LINKED ISSUE #1808: Adopt docking window style based on content panel framework

bevans2000:
This focuses on improving the layout of the Docking window approach. Highlights are:
- Docking Windows can be reopened after closing
- Changed layout of AttributePanel in dynamic mode
- Prevent Docking framework adding an extra scroll pane
- Some Tab Panels adjusted to improve layout

In addition there is now a '-debugui' command argument to switch on size displaying and the Entity Listener panel

Close #1808

### DIFF
--- a/mars-sim-core/src/main/resources/messages.properties
+++ b/mars-sim-core/src/main/resources/messages.properties
@@ -77,6 +77,7 @@ mission.designation=Designation
 person.singular=Person
 person.plural=People
 person.birthDate=Birth Date
+person.age=Age
 person.bloodType=Blood Type
 person.emotion=Emotion
 person.energy=Energy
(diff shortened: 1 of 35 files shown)


bevans2000:
@mokun I think the first pass of the Docking approach is done with this PR. The rewrote of the Swing content has been very worthwhile as it's all a lot cleaned and simplified. The windowing approach can be changed from the Settings menu now.
The restored layout for Docking can't handle if the user does complex layout but it restores to a best estimate.
Give it a try and tell me what you think

mokun (approved):

PR #1919, opened 2026-05-02
fix(ui): Improve filtering on EventViewer

### LINKED ISSUE #1918: Too many Event in EventVIewer
**Describe the bug**
On the default Scenarios where there are  9+ Settlement there are too many Events shown in the Event Viewer. 
Whilst Category filtering helps it maybe the user wants to "follow" a Settlement.

Add an additional filter to allow the user to select by Settlement or Authority. This would have the same selection logic as the Monitor Window.

bevans2000:
Changes:
- Extract Authority/Settlement selector in MonitorWindow to SettlementsSelector
- SettlementsSelector allows use to select either Settlement or Authority but presents a list of SettlementsSelector
- Update MonitorWindow to use seperate selector
- Add a SettlementsSelector to the EventViewer toolbar, but redesign
- EventViewer filters on Settlement as well as Category
- Add a Toggle All option to the category filter

close #1918

### DIFF
--- a/mars-sim-ui/src/main/java/com/mars_sim/ui/swing/tool/eventviewer/EventViewer.java
+++ b/mars-sim-ui/src/main/java/com/mars_sim/ui/swing/tool/eventviewer/EventViewer.java
@@ -84,6 +84,7 @@ public class EventViewer extends ContentPanel implements ConfigurableWindow, His
         loadInitialFilter(userSettings);
 
         initializeUI(userSettings);
+        selectedSettlements = settlementSelector.getSelectedSettlements();
         loadEvents();
         
         // Register as listener for new events
@@ -340,6 +341,7 @@ public class EventViewer extends ContentPanel implements ConfigurableWindow, His
         if (eventManager != null) {
             eventManager.removeListener(this);
         }
+        settlementSelector.unregister();
         super.destroy();
     }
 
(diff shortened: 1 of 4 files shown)


mokun (approved):

mokun (commented):
@bevans2000 

In the past, the settlement/authority combobox has a very wide textfield. The internal width of the box always exceeds the length of the name of the settlement/authority 2x or 3x times. 

Can we limit the width to, say, no more than 120% of the length of the longest name ?

bevans2000:
> @bevans2000 
> 
> In the past, the settlement/authority combobox has a very wide textfield. The internal width of the box always exceeds the length of the name of the settlement/authority 2x or 3x times. 
> 
> Can we limit the width to, say, no more than 120% of the length of the longest name ?  

I have already removed the explicit size setting on the combo box so it will be smallet now

mokun:
Right now, in Monitor Tool, the combobox would resize by itself from time to time in the same session when I click on it to get another settlement/sponsor.

Has you seen that ?

Sometimes it would make the width is small that it truncates the name of the settlement being selected.

Does resizing the Monitor Tool window affects its width ?

Note that I mainly uses it in Settlements tab and Buildings tab.

mokun:
It seems whenever the 'All" was chosen, it resizes and reduces the width and it causes the settlement name to be truncated afterward

See pic below : 

Screenshot 2026-05-30 at 19 41 31

bevans2000:
It will be a sizing thing. I'll take a look

Issue #1340, opened 2024-07-07
Create a Metrics Manager


bevans2000:
**Is your feature request related to a problem? Please describe.**
In many separate places metric over time are captured using a variety of classes. The individual implementations have different retention periods and data storage.
Viewing the metrics is custom to each one, needing bespoke code in the UI.
Furthermore there is no ability to retain the data for long periods. 
Relates to #973

**Describe the rationale**
Metrics is key to understand how the Simulation and Settlement are performing.

**Describe the goal(s) you have in mind**
Consolidate the various metrics into a single solution that provides a consist and universal way to store any numerical metric. Having a single solution will lead to a sharable means to render the metrics using tabular or chart based presentations. These could show changes over time.
A central solution will also allow better management of the storage of these metrics which may become large

**Describe the solution you desire to see**
Create a new ```MetricManager``` class that is the central point for holding metrics. An instance is created per Settlement which removes the problem of multiple updaters. This will replace all '''SolLogger''' classes.
The Metric manager will capture absolute value of a named metric at a point in time. It will be able to retrieve the sequence of metrics over time. Metrics will be identified by 2 attributes:

1.  Type - This could be Resources Consumed, Active Missions, EVA hours
2.  Instance - This will be a sub-type and specific to the Type. For example, name of resource for Consumed, Type of Mission or Person name

The attributes will be String based to allow for expansion.
The Metric Manager will also handle storage by writing old sols to file as CSV ? to control the amount of memory consumed.
Older data will automatically be loaded from file store if it is required; this will use the Future concept.

To accompany this, a central UI will be provided that can chart any selected metric over time allowing metrics from different type to be rendered side by side.

mokun:
Very good idea to have a unified Metric Manager that can output to CSV at player's demand by pressing a button on a UI or simply on console. 

btw, it's actually good for us developers to be able to do a data dump on, say, the production or consumption of a methane or any resources; or the power/heat consumed or generated at a power/heat source of building, etc.

bevans2000:
It wasn't in my original idea to have an export to CSV but we can add it.
The idea was to use CSVs to hold old data so less is kept in memory.

mokun:
How do you like to handle saving and loading from external files ?

Creating a default.csv or default.json for the data ?

bevans2000:
[...]

PR #1927, opened 2026-05-10
feat: Create unique means to identify any Entity externally

### LINKED ISSUE #1905: Create an EntityIdentifer
**Is your improvement request related to a situation? **
To be able to store Metric in a database the `Metric` class must be storable in a database and when retrieved must be mapped back to the objects in memory. The problem is the `Entity` reference is an interface that could map to any number of objects.

**Describe goal(s) you have in mind**
The solution is to create a `EntityIdentifier` class that unique identified any Entity. As all Entity objects do not have to be global objects, some are controlled by a Settlement. The identifier must have a reference to the parent as well as the type and identifier of the object itself.

The identifier will need the ability to stringify so that it can be transferred external, either database, file or API. 

A `EntityResolver` will convert from the identifier back into the appropriate Entity objects.

**Future Use**
In the future the `EntityIdentifer` can be used in an API to support a Web or mobile client; essentially becoming the Noun in a RestAPI.

bevans2000:
This PR creates an `EntityIdentifier` that is generated by any `Entity`.  A new `EntityResolver` class can resolve a previously create identifier into the relevant Entity.
This has been adopted in the EntityContentFactory and will support #1340.

close #1905


mars-sim-core/src/main/java/com/mars_sim/core/EntityResolver.java
```diff
@@ -0,0 +1,53 @@
...
+ * This is a static helper class.
+ */
+public class EntityResolver {
+
+    private EntityResolver() {
+        // Private constructor to prevent instantiation
+    }
+
+    public static Entity resolve(Simulation simulation, EntityIdentifier identifier) {
+        return switch(identifier.type()) {
```
mokun, line 20: Awesome that many types of objects (even if they don't extend Unit) will soon be defined as an Entity !

bevans2000: Glad you like it. I'm sure I've got it working but just checking it out.
We can use this when I do a GraphQL interface because it gives a consistent external reference. 
But the main short term use is for the metrics manager so i can store all the metrics in a database. There will be too many to hold in memory. This might be the start of dropping serialisation and going to database although that is a long journey.
It's now use it saving the window layout to hold the reference for any EntityContent window.

mokun (approved):

mokun:
So when it comes to airlock operations, do you see anyone getting stuck and starved or dehydrated now ?

Btw, I'm still not sure if the teleportation issues have been resolved. I've added detail log for tracking what the location state of a person is if it happens.

I'll run some test tomorrow morning.

bevans2000:
Running some tests would be good.
But I'm now thinking we should rewrite the airlock logic from scratch using a state based machine approach along with the associated Tasks. A clean approach might solve some of the complexity.
I will push this change tonight.

mokun:
>  using a state based machine approach along with the associated Tasks. 

The thing is that we have too many subtasks currently within EnterAirlock and ExitAirlock.

We should be able to break down these two monolithic tasks into, say, 6 simpler tasks.

And it's not difficult to see that we can even reuse these tasks in other scenarios.

mokun:
FindAirlock would be the first task.

The result of this task will be true if an airlock chamber is available. Or false if chamber is not available.

If a person is inside and a chamber is available, it leads to the next task called FindInnerDoor.

If a person is outside and a chamber is available, the next task is FindOuterDoor.

And so on.

bevans2000:
I'll create an issue for it.
But I think a single Tasks for ingress and egress make better sense. We need clear use of Task phase.
One thought on finding an airlock. If we store EVA suits in the airlock building then a Person will also go the Building holding their assigned suit

PR #1940, opened 2026-05-23
feature: Settlement Supplies states the types of Robots delivered.

### LINKED ISSUE #1883: SettlementTemplate and Settlement Supplies should define Robot types
This is a follow-on from #1880.
The Settlement template and resupply missions define a number of robots with the simulation deciding what types are needed. In the case if a newly arriving Settlement or Resupply the types of Robots would have to be defined upfront before the ship departed.

The `SettlementSupplies` should support a manifest of Robots of certain types. The impact will also be on the initial settlement configuration in a Scenario. A number of Robots will no longer be sufficient; the types have to be pre-selected.
It will allow the removal of the automatic Robot selection logic.
There will be a minor change on the UI but the majority will be on the core engine and configs.

In addition the contents of a Settlement Supply should be customisable once assigned to a Transport Mission

bevans2000:
This reworks how supplies are defined and makes them more isolated for better reuse. Following the example below in the settlement template specifications

```xml
<initial-supplies>
   <robot type="Medic-Standard" number="2"/>
   <vehicle type="Rover" number="1"/>
</initial-supplies>
```

The robots are also now included in the supplies as Transport mission must have a pre-defined cargo before embarking on the journey; the use of the auto allocation algorithm is incorrect.

The helpGenerator and Scenario Editor have been updated to reflect this re-design.

close #1883


mokun:
> pre-defined cargo 

You mentioned cargo. That makes me think of physical container box that package a merchandise. 

So currently, are cargo delivery good between two settlements ?

The logistic business of sending and receiving goods can be privatized as settlements go beyond the survival phase.

And low density packaging material may need to be manufactured on Mars as well

mokun (approved):

bevans2000 (commented):
> 
> So currently, are cargo delivery good between two settlements ?
> 
This is focused in `SettlementSupplies` which is used for Transports and building Settlement. For the Transport missions I think we can assume that teh ship has the ability to store the goods in the manifest (SettlementSupplies).
This was solely aimed at removing the oiutliner of how Robots are defined for a transport mission.

Issue #1945, opened 2026-05-26
Unable to back step in StartUpChooser


mokun:
**Describe the bug**
- Currently, a player can highlight one of the items in the Main Menu of StartUpChooser and click that choice to proceed. 
- However, upon seeing a dialog of that choice, he closes the dialog box and would like to go back to the Main menu. The Main Menu would no longer work and won't give player to choose another item again. 
- Also, the close button of StartUpChooser does not work and thus unable to terminate StartUpChooser.

**Expected and unexpected behavior**
- If the dialog was closed, it is assumed that the player would like to go back to a functional Main Menu to choose another item.

**Java classes involved**
- StartUpChooser

**Screenshot**
- None of the items can be highlighted again upon closing a dialog box.
- The close button of StartUpChooser does not work either.
<img width="295" height="361" alt="Image" src="https://github.com/user-attachments/assets/039dd882-457c-405c-9059-f8dfe0fce4c5" />

**Specifications**
- pre-4.1

**Note**
- It is problematic to resort to calling choiceMade(EXIT) from time to time. 

```java
        addWindowListener(new WindowAdapter() {
			@Override
			public void windowClosing(WindowEvent event) {
				choiceMade(EXIT);
			}
		});
```

- Calling choiceMade(EXIT) if no sim file is selected is not good. Need to simply go back to a functional main menu.

```java

    /**
	 * Performs the process of loading a simulation.
	 */
	private void selectSimFile() {
		JFileChooser chooser = new JFileChooser(SimulationRuntime.getSaveDir());
        chooser.setFileFilter(new SimFileFilter());
		chooser.setDialogTitle(Msg.getString("MainWindow.dialogLoadSavedSim")); // -NLS-1$
		if (chooser.showOpenDialog(this) == JFileChooser.APPROVE_OPTION) {
			selectedFile = chooser.getSelectedFile().getAbsolutePath();
            choiceMade(LOAD_SIM);
		}
        else {
        	choiceMade(EXIT);
        }
	}

```

mokun:
@bevans2000 

Can you try fixing that ?

bevans2000:
@mokun  I've used the new Priority field as Urgent

mokun:
> new Priority field as Urgent

meaning ?

bevans2000:
It's a public preview of new feature https://github.com/orgs/community/discussions/189141

mokun:
Just tested it on window 11.

Same problem as on my macOS.

[...]

PR #1954, opened 2026-05-30
fix(ui): Selectiondialog forces CANCEL on window exit

bevans2000:
ref #1945

### DIFF
--- a/mars-sim-ui/src/main/java/com/mars_sim/ui/swing/main/SelectionDialog.java
+++ b/mars-sim-ui/src/main/java/com/mars_sim/ui/swing/main/SelectionDialog.java
@@ -11,6 +11,8 @@ import java.awt.Container;
 import java.awt.Dialog;
 import java.awt.Dimension;
 import java.awt.event.ItemEvent;
+import java.awt.event.WindowAdapter;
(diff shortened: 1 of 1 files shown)


mars-sim-ui/src/main/java/com/mars_sim/ui/swing/main/SelectionDialog.java
```diff
@@ -130,6 +132,15 @@ private SelectionDialog(Dialog owner, JPanel mainPane, String title) {
  
         //Initialize values.
         pack();
+
+        // Closed via toolbar then cancel
+        addWindowListener(new WindowAdapter() {
+			@Override
+			public void windowClosing(WindowEvent event) {
+				choiceMade(CANCEL);
```
mokun, line 140: I see. Good about rerouting it to calling cancel.

mokun (approved):

mokun:
We should do the same to make Simulation Configuration Editor (SCE) fall back to startup chooser if player click the top right exit button.

But currently SCE lacks a cancel button. Should we simply add another dialog box to ask player to confirm exiting SCE ?

PR #1973, opened 2026-06-20
feat: Use Mertic to record Mission stats

bevans2000:
Changes:
- Added MetricGroup to record Metric Measure that share the same Category and Entity
- Add Unit test
- Completed missions notifiy parent MissionControl of the completion
- MissionControl replaces SolHistory with MetricGroup
- MissionControl record approvals, rejected, succesful, aborted and started missions
- Updated MissionPlanCommand to display MetricGroup stats
- Simulation uses MemoryMetricManager with a 10 sol limit

DatabaseMetricManager is too unstable tobe default currently

ref: #1340

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,7 @@
 **/target/
 .classpath
 .project
-.settings/
+**/.settings/
 .idea
 .prefs
 /mars-sim.iml
--- a/mars-sim-console/src/main/java/com/mars_sim/console/chat/simcommand/settlement/MissionPlanCommand.java
+++ b/mars-sim-console/src/main/java/com/mars_sim/console/chat/simcommand/settlement/MissionPlanCommand.java
@@ -7,12 +7,12 @@
 
 package com.mars_sim.console.chat.simcommand.settlement;
 
+import java.util.HashMap;
 import java.util.Map;
 
 import com.mars_sim.console.chat.ChatCommand;
 import com.mars_sim.console.chat.Conversation;
 import com.mars_sim.console.chat.simcommand.StructuredResponse;
-import com.mars_sim.core.person.ai.mission.PlanType;
 import com.mars_sim.core.structure.Settlement;
 
 /**
@@ -43,7 +43,7 @@ public class MissionPlanCommand extends AbstractSettlementCommand {
 		context.println(prompt2);
 		int newObj = context.getIntInput("Select an option ");
 		int max = 0;
-		boolean totals = false;
+		Map<String, Double> totals = null;
 		int today = context.getSim().getMasterClock().getMarsTime().getMissionSol();
 
 		switch (newObj) {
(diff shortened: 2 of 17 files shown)


bevans2000 (commented):
@mokun This PR records all Mission stats for any Settlement. Run the simulatino with multiple Settlemtns and then use the Metric View to compare the stats in charts, either in one chart with multiple lines or multiple charts

bevans2000:
@mokun Don't merge this as it's still a Draft. If it's Draft don't merge PRs because they may be broken.
Just trying to fix something the Simulation using a H2 database by default.

mokun:
> Don't merge this as it's still a Draft

Sure.

bevans2000:
I've uncovered a problem with the H2 database not resetting sometimes.
So tomorrow I will complete this change to get full stats on Missions but still use the in memory metrics until I can figure it out. This will allow you to have a go trying to capture some metrics yourself.
In the meantime, I will implement a Sol limit on the Memory metric manager so it does run away with memory. That's quite easy

mokun (approved):

mokun:
After you merge the change, I'll play around it.

Is there any particular reason that at this stage, you want to capture mission stats right away ?

Why not just the resources of, say, a greenhouse ?

bevans2000:
No serious reason. We capture the mission approvals already so I want to replace that.

bevans2000:
@mokun So I've have merged this in. If you run the Simulation for a few Sols then the Metric Viewer should record 4 Measures in the Mission category for each Settlement. It is currently held in memory so I've put a limit of 10 Sols of history. You can change this if you want.
Look at the MissionControl use of MetricGroup as a good example. So for your Greenhouse scenario I would create a new MetricCategory, e.g Farming and record the various values as different Measures. You could record these against the Settlement or Greenhouse Building. 
We might need to think how the Entity name is rendered in the Chart key as it just uses Entity name currently. But as it's Entity we could add the context attribute to help if there are 2 Greenhouses the same name.
Give it a try.
I'll work on the database version to allow us to store a lot of metrics.

mokun:
> as it's Entity we could add the context attribute to help if there are 2 Greenhouses the same name.

Do you mean players may pick two greenhouse from two different settlement for comparison?

Hmm I havent thought of the need to handle this collision possibility until now.

Do we want to make unique name and/or globally for each building soon ?

Instead of having two "Lander Hab 1"'s, we can be more creative with the building names.

Say we can say it in its native language to signify the country it belongs.

bevans2000:
So I believe there is only a need for globally unique names in Settlements and Scientific Study. Those are the only Entities that are global, everything else, Building, Robot are owned by Settlement. Basically we already have this pattern in the Entity interface with the context property.
So yes, why wouldn't a user want to compare 2 Greenhouses from 2 different Settlements.
I think we could do something in the Entity drop-down to show the context in the future.

Issue #1986, opened 2026-06-28
Isolate Mission creation from Running the mission


bevans2000:
The current structure has methods that help select the members inside the Mission subclasses. When moving towards the newer MissionProject approach there will not be concrete subclasses of Mission; they are all generic with the combination of steps defining the purpose. The parameters for selecting Members means a Mission has to be created before members can be identified potentially resulting in the creation of Missions that instantly fail.

A better solution would be to put the members selection parameters into the MetaMission allowing members to be identified before a Mission. Also for adhoc missions in the future a bespoke on the fly MetaMisison could be created to capture the unique selection parameters for the combination of mission steps.
This would then lead to extracting the mission recruitment logic into a new ```MissionBuilder``` class that looks to recruitment members for a MetaMission and preselected Leader. The recruitment logic could work on sub-populations of a Settlement, e.g. settlers or tourists.

Whilst simpler the same argument applies to vehicle selection. This should be done outside before the Mission is created based on MetaMission attributes.

bevans2000:
I think Meta mission should have the following behaviour

- Define the minimum and default maximum members
- Define the preferred job types for worker members to join
- Define the preferred Robot types for members to join
- Rate the job types for the potential leaders, this is not a list because should jobs are better than other for leading, e.g.  rescue mission is best lewd by a Technician but a Engineer or Pilot will do
- Calculate the mission qualification score

At the moment all the above are scattered about the existing Mission logic so this will bring into one place. Then in the future these attributes can be created automatically depending on the proposed mission steps hence supporting ad-hoc missions.

This sounds good and right BUT it's a lot of moving things around. Probably need to write UnitTest for each different Meta mission to ensure the new logic works correctly.

mokun:
> A better solution would be to put the members selection parameters into the MetaMission allowing members to be identified before a Mission.

Q: Is it true that MetaMission is created 50x more than Mission ? 

If that's the case, will the preselection of members end up more negatively impacting the performance of the simulation ?

mokun:
>  extracting the mission recruitment logic into a new MissionRecruitment class that looks to recruitment members for a MetaMission and preselected Leader.

So can we have at least ONE or TWO or MORE combinations of groups (depending on the population) of members already being preselected ahead of time ? 

Now I see making this recruitment process being independent of any potential missions will save a lot of repetitive work of finding members needlessly.

[...]

PR #1990, opened 2026-07-04
feat:  MetaMission controls thresholds for Mission creation

bevans2000:
This PR moves the control of any Preferences relating to Missions under the Mission Control class. 
In particular a new Preference is added that allows the user to disable the Sol threshold for creating missions. This is useful for testing o running an accelerated simulation.

To support this, MetaMission now holds the minimum sol threshold for each mission as a explicit property. It is no longer embedded in the probability calculations. The new property is used by the mission control when searching for new missions.

The meta mission is also responsible for calculating how many concurrent missions a population count can support via the getMaxMissions method. This uses a combination of a population threshold, min people before mission can be created, and a population ratio, govern additional missions.

This is part of #1986


mokun (approved):

bevans2000:
@mokun Also we have over 1000 unit tests now with this PR

mokun:
> over 1000 unit tests now with this PR

So you mean we have altogether 1000+ int tests now in total ?

How do you count them ?

bevans2000:
It shows when I run the tests in my IDE

PR #2012, opened 2026-07-18
fix: Resources collected recorded at site

### LINKED ISSUE #2002: Incorrect Site Name
**Describe the bug**
- On a CollectIce or CollectRegolith mission, when parsing the site number or name, something is not quite right. 

**Suggestions for future work**
- Consider caching a list of names of resource collection/exploration/mining sites. 
- Instead of generating new sites each time, Reuse that same list of sites and give meaningful names, much like in many of those NASA Mars surface missions.
- Download a list of site names that NASA uses or generate our own.

**Java classes involved**
- I traced the issue to 'TabPanelObjectives` that create `CollectResourcePanel`.

In `CollectResourcePanel`,
```
/**
	 * Updates the collection value label.
	 */
	private void updateCollectionValueLabel() {
		resourceModel.clear();
		objective.getResourcesCollected().entrySet().stream()
			.map(e -> ResourceUtil.findAmountResourceName(e.getKey()) + ": " + StyleManager.DECIMAL_KG.format(e.getValue()))
			.sorted()
			.forEach(resourceModel::addElement);

		siteModel.clear();
		objective.getCollectedAtSites().entrySet().stream()
			.map(e -> "Site #" + Integer.toString(e.getKey()) + ": " + StyleManager.DECIMAL_KG.format(e.getValue()))
			.sorted()
			.forEach(siteModel::addElement);
	}

```

The line ` "Site #" + Integer.toString(e.getKey())` is problematic. 

Q: Should we stick with the Navpoint number designation or the site name (e.g. `Prospecting Site #2`) ?

I don't why e.getKey() in Integer.toString() will return "-1".

I look up a number of these resource collect missions.

In the Objectives tab, they all show "-1" as the first site. Then they jump to `Site #1` and so on. 

In `CollectResourcesMission`,

```java
	/**
	 * Records the amount of resources collected.
	 * 
	 * @param resourceType
	 * @param samplesCollected
	 */
	public void recordResourceCollected(int resourceType, double samplesCollected) {
		// Update amountCollectedBySite
		int siteIndex = getCurrentNavpointIndex(); 
		objective.recordResourceCollected(siteIndex, resourceType, samplesCollected);
	}

```

It jumps to `AbstractVehicleMission`,
```java
	/**
	 * Gets the index of the current navpoint the mission is stopped at.
	 * 
	 * @return index of current navpoint or -1 if mission is not stopped at a
	 *         navpoint.
	 */
	public final int getCurrentNavpointIndex() {
		if (travelStatus != null && AT_NAVPOINT.equals(travelStatus))
			return navIndex;
		else
			return -1;
	}
```

This is where "-1" is being returned.

**Screenshots**
Image

**Specifications**
- pre-4.1

**Related Issues and Discussions**
- https://github.com/mars-sim/mars-sim/issues/1968#issuecomment-4951819193

bevans2000:
Changes:
- CollectResourceObjectives record resources collected against site name
- Resources panel uses site name instead of generating
- CollectResourcesMission records current site when new phase starts
- CollectResourcesMission uses current site to record resources collected
- Replace objectives icon with trophy

close #2002


mokun:
In terms of naming a site, how creative can we get ? Shall we put together a list of site names and auto generate a xml or json file to hold them ?

bevans2000:
The point is for the sites for resources collection are randomlt lynchoosen so there is no point using a predefined name

mokun (approved):

PR #2026, opened 2026-07-25
fix: Add more defensive coding to handle concurrency problems

bevans2000:
Changes:
- RetryRowSorter now applies the catch/retry pattern when a resort is triggered via a table event
- Disable auto resorting on an table event change
- SettlementSelector uses the AWT Event thread to change the selection. This avoids the Unit Threads updating the UI
- Simulation class uses the try/catch pattern when applying the pulse to top level temporals; This reduces the blast radius of failures


mokun (approved):

mokun:
May be related to https://github.com/mars-sim/mars-sim/commit/ebf70bf830e7da3eb2e1e2be380288d5f9776c61#commitcomment-193765057.

bevans2000:
@mokun Auto sorting is now disabled, try it again

mokun:
Do you mean fully disabled?

bevans2000:
Yes, the table will update but it will not auto-sort by default. We need to revisit this but for now let's make sure it works in your cases. I believe auto sorting on event was disabled in the default table sorter.

Issue #2022, opened 2026-07-24
Null initialWalkState and unable to return to the settlement


mokun:
**Describe the bug**
- Recently, mining missions have stranded settlers on the surface of Mars.
- The mining mission vehicle went out of resources.
- They were unable to return to a settlement or vehicle due to invalid walking state.
- They were having high fatigue as a result.
- Excessive number of `initialWalkState : null`results.

**Java classes involved**
- `WalkingSteps`

**Stacktrace**

```
01-Adir-05:055.346(Info) MissionControl : [Heritage] Mining 4-9 - Review completed, outcome is Approved, score: 1096.4 [Min: 102.1].
...
01-Adir-05:063.420(Info) AbstractMission : [Heritage] Mining 4-9 - Mission plan approved.
...
01-Adir-05:505.366(Warning) AbstractVehicleMission : [14.9570 N 114.0660 E] Sand Spirit - Not enough Water to continue with Mining 4-9 - Required: 77.76 kg - Vehicle stored: 67.69 kg.
01-Adir-05:505.366(Info) AbstractVehicleMission : [14.9570 N 114.0660 E] Sand Spirit - Reported no lacking of resources. Turning toward Heritage. New distance: 11.161
01-Adir-05:505.366(Info) EVAMission : [14.9570 N 114.0660 E] Sand Spirit - EVA ended due to external trigger.
01-Adir-05:505.366(Info) EVAMission : [Mars Surface] Hanako Li - EVA operation Terminated
01-Adir-05:505.366(Severe) AbstractVehicleMission : Mining Site's newNavIndex 3 is out of bounds.
01-Adir-05:505.366(Info) AbstractVehicleMission : [15.1255 N 114.1535 E] Sand Spirit - Arrived back home Heritage.
01-Adir-05:505.366(Severe) Vehicle : [15.1255 N 114.1535 E] Sand Spirit - Not found to be parked in a settlement.
01-Adir-05:505.366(Info) RoverMission : [Heritage] Sand Spirit - Disembarked at Heritage triggered by Hanako Li.
...
01-Adir-05:511.173(Info) AbstractMission : [Mars Surface] Hanako Li - Ended the Mining 4-9 with status flag(s): Not enough Water.
01-Adir-05:511.173(Info) AbstractMission : [Mars Surface] Hanako Li - Disbanding mission member(s): Hanako Li, David Huber
[...]

PR #2030, opened 2026-07-25
fix: Prevent EVA continuing when mission aborted

bevans2000:
This is not guaranteed to fix tbhe problem but aims to gracefully handle the situation when an EVA is aborted.

Changes:
- In EVAMission stop new EVAs once EVA is cancelled
- When aborting return direct to the home settlement

ref #2022

### DIFF
--- a/mars-sim-core/src/main/java/com/mars_sim/core/person/ai/mission/AbstractVehicleMission.java
+++ b/mars-sim-core/src/main/java/com/mars_sim/core/person/ai/mission/AbstractVehicleMission.java
@@ -1116,67 +1116,6 @@ public abstract class AbstractVehicleMission extends AbstractMission implements 
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


mokun (approved):

mokun:
So this is the stop gap solution for fixing the EVA. Hope the new mission project EVA won't need to use EVAMission class at all. Will it ?

bevans2000:
> So this is the stop gap solution for fixing the EVA. Hope the new mission project EVA won't need to use EVAMission class at all. Will it ?


Yes it's just a stop gap. I couldn't totally follow how the NavPoints are changed when a mission is aborted, I think this a problem. But this change should at least force everyone back to the river before returning home.
My hope is the new way will significantly simplify this scenario. Note currently the new approach does check for a lack of resources but it should be easier to add such things because it is more modular.

PR #2032, opened 2026-07-26
Fix construction mission LUV handling

### LINKED ISSUE #2028: ConstructionMission incorrectly references GroundVehcile
**Describe the bug**
The `ConstructionMission` accepts a list of `GroundVehicle` but the internals use LUV everywhere. Also the Mission  Wizard does not allow the use to select ANY vehicles to help.

** Solution **

Change the constructor and internal properties to be based on LUV. Add a new step to the Mission wizard to select multiple LUVs and add them to any created Construction Mission.

FSL-house:
## Summary

- Use `LightUtilityVehicle` directly for construction mission vehicle collections.
- Remove unnecessary `GroundVehicle` typing and casts around construction vehicle handling.
- Allow the mission wizard's LUV step to select multiple LUVs for construction missions while keeping mining missions single-select.
- Show selected construction LUVs on the wizard confirmation panel.
- Add a regression test that keeps `ConstructionMission` construction vehicles typed as `LightUtilityVehicle`.

## Why

Construction missions reserve and operate light utility vehicles specifically. Keeping the public mission/objective data as generic `GroundVehicle` makes the API less precise and forces casts in construction task code.

Fixes #2028.

## Validation

- `mvn -pl mars-sim-core -Dtest=ConstructionMissionTest test`
- `mvn -pl mars-sim-ui -am -DskipTests compile`

### DIFF
--- a/mars-sim-ui/src/main/java/com/mars_sim/ui/swing/tool/missionwizard/ConfirmationPanel.java
+++ b/mars-sim-ui/src/main/java/com/mars_sim/ui/swing/tool/missionwizard/ConfirmationPanel.java
@@ -53,7 +53,7 @@ class ConfirmationPanel extends WizardStep<MissionDataBean>  {
             attrs.addLabelledItem(Msg.getString("lightutilityvehicle.singular"), new EntityLabel(state.getLUV(), context), null);
         }
         if ((state.getConstructionVehicles() != null) && !state.getConstructionVehicles().isEmpty()) {
-            attrs.addLabelledItem(Msg.getString("lightutilityvehicle.singular"),
+            attrs.addLabelledItem(Msg.getString("lightutilityvehicle.plural"),
                     createEntityList(state.getConstructionVehicles(), context), null);
         }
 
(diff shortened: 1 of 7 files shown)


mars-sim-ui/src/main/java/com/mars_sim/ui/swing/tool/missionwizard/ConfirmationPanel.java
```diff
@@ -52,6 +52,10 @@ class ConfirmationPanel extends WizardStep<MissionDataBean>  {
         if (state.getLUV() != null) {
             attrs.addLabelledItem(Msg.getString("lightutilityvehicle.singular"), new EntityLabel(state.getLUV(), context), null);
         }
+        if ((state.getConstructionVehicles() != null) && !state.getConstructionVehicles().isEmpty()) {
+            attrs.addLabelledItem(Msg.getString("lightutilityvehicle.singular"),
```
bevans2000: This should use the new plural derivative

bevans2000 (changes requested):
We follow the Conventual Commit standard. Could the summary be changed to fix:..…
But the PR looks good

mars-sim-ui/src/main/java/com/mars_sim/ui/swing/tool/missionwizard/ConfirmationPanel.java
FSL-house: Thanks! Updated the confirmation panel to use the plural LUV label and changed the follow-up commit to use Conventional Commit format. I also updated the PR title to `fix: handle construction mission LUVs`.😊

bevans2000: Great, I've merged it in.
Thanks for contributing. Feel free to jump in on other issues if you want

FSL-house: Thanks for the review and quick merge! I’ll look for another issue to work on.

PR #2044, opened 2026-08-10
fix(core): Fix building area calculation by adding shape attribute in buildings.xml

### LINKED ISSUE #2031: Building names used in Building set-up
**Problem**
The Building constructor search for a specification with name 'Hab' to change how the building area is calculated. It assumes if the building spec name contain "Hab" it is round. This is unstable logic.
Other building maybe round and some type of Hab could be rectangular.

**Solution**
The  area covered by a Building should be defined in the BuildingSpec. This area is fixed up front by either:

1. Defined in the building.xml file
2. Shape (rectangular or circular) defined in building.xml and then calculated.
3. No shape defined; then assume rectangular

The Building constructor will take the area directly off the spec and remove the dependency on the buildingspec name.

Shalinihr1402:
Fixes #2031

Re-opening this PR against the official repository (my previous PR was accidentally opened against a personal fork with an outdated base).

Changes:

Added a shape="circular" attribute to buildings.xml for all circular Habs and Hubs.
Removed the hardcoded string matching (name.contains("Hab")) in Building.java.
BuildingConfig.java now parses the shape and calculates exact circular floor area using pi * r^2.
LifeSupport.java now relies on this exact floor area from the BuildingSpec instead of overestimating with length * width.
This should completely fix the oxygen and heat overestimations for the circular dome buildings.

### DIFF
--- a/mars-sim-core/src/main/java/com/mars_sim/core/building/Building.java
+++ b/mars-sim-core/src/main/java/com/mars_sim/core/building/Building.java
@@ -123,7 +123,6 @@ public class Building extends FixedUnit implements Malfunctionable,
 	// Regarding the length, how to best handle the variable length of brickway/walkly/hallway/tunnel 
 	private double length;
 	private double floorArea;
-	private String shape;
 	private double areaFactor;
 	private double facing;
 
@@ -214,7 +213,6 @@ public class Building extends FixedUnit implements Malfunctionable,
 		
 		this.floorArea = buildingSpec.getArea();
 		this.areaFactor = Math.sqrt(this.floorArea) / 2;
-		this.shape = buildingSpec.getShape();
 		
 		constructionType = buildingSpec.getConstruction();
 
@@ -1420,10 +1418,6 @@ public class Building extends FixedUnit implements Malfunctionable,
 
 	// TODO this is wrong as names can change. This is just used to identify if there are multiple floors.
 	public boolean isAHabOrHub() {
-		if (shape != null) {
-			return "circular".equalsIgnoreCase(shape);
-		}
-		// Fallback for tests if shape is not loaded
         return buildingType.contains(" Hab") || buildingType.contains(" Hub");
     }
 
--- a/mars-sim-core/src/main/java/com/mars_sim/core/building/Building.java
+++ b/mars-sim-core/src/main/java/com/mars_sim/core/building/Building.java
@@ -123,6 +123,7 @@ public class Building extends FixedUnit implements Malfunctionable,
 	// Regarding the length, how to best handle the variable length of brickway/walkly/hallway/tunnel 
 	private double length;
 	private double floorArea;
+	private String shape;
 	private double areaFactor;
 	private double facing;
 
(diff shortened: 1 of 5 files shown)


mars-sim-core/src/main/java/com/mars_sim/core/building/Building.java
```diff
@@ -1420,8 +1420,11 @@ public double getCurrentAirPressure() {
 
 	// TODO this is wrong as names can change. This is just used to identify if there are multiple floors.
 	public boolean isAHabOrHub() {
-        return buildingType.contains(" Hab")
-                || buildingType.contains(" Hub");
+		if (shape != null) {
```
bevans2000: Don't hold the shape as a field. Just use the name as a check here.
This method needs reworking as it's bad being dependent on the building type name. But that is for another change

mars-sim-core/src/main/java/com/mars_sim/core/building/Building.java
```diff
@@ -216,6 +212,10 @@ public Building(Settlement owner, String id, int zone, String name, BoundedObjec
 						BuildingSpec buildingSpec) {
 		this(owner, id, zone, name, buildingSpec.getValidBounds(bounds), buildingSpec.getName(), buildingSpec.getCategory());
 		
+		this.floorArea = buildingSpec.getArea();
+		this.areaFactor = Math.sqrt(this.floorArea) / 2;
+		this.shape = buildingSpec.getShape();
```
bevans2000: As below, don't save this.

bevans2000 (changes requested):
Can you change the name of the PR so it matches Conventional Commit style please. So this would be fix(core):

bevans2000:
@Shalinihr1402  closed and merged. 
Thanks for the contribution, feel free to look at any of the other unassigned issues

Shalinihr1402:
@bevans2000 Thank you! This was my very first open-source contribution, so thank you for the help and the opportunity. I'll definitely check out the other issues!

mokun:
After this commit, the Heating class keeps throwing error when it comes to Hallway.

```
01-Adir-01:002.872(Severe) Heating : [Star Base] H10 - selReqHeat is NaN.
01-Adir-01:002.872(Severe) Heating : [Star Base] H11 - selReqHeat is NaN.
01-Adir-01:002.872(Severe) Heating : [Star Base] H9 - selReqHeat is NaN.
01-Adir-01:002.872(Warning) ThermalGeneration : [Terminus] T1 - 3. heatGen: 0.0  finalHeatReq: 22.76  T: 0.0  devT: 18.0  mod: 2.8  mHeatReq: 19.6  heatReq: 0.0  remainHeatReq: 22.76  preNetHeat: -0.63  postNetHeat: -0.63  millisols: 0.957
01-Adir-01:003.829(Severe) Heating : [Heritage] H18 - activeVentHeat is NaN.
01-Adir-01:003.829(Severe) Heating : [Korolev] H13 - activeVentHeat is NaN.
01-Adir-01:003.829(Severe) Heating : [Korolev] H14 - activeVentHeat is NaN.
01-Adir-01:003.829(Severe) Heating : [Korolev] H12 - activeVentHeat is NaN.
01-Adir-01:003.829(Warning) Heating : [Terminus] H1 - I: 0.0  canopyFactor: 0.0  canopyHeatGain: 0.0  emissivity: 0.251  solarHeatLoss: 0.0
01-Adir-01:003.829(Severe) Heating : [Terminus] T1 - activeVentHeat is NaN.
01-Adir-01:003.829(Severe) Heating : [Heritage] H17 - activeVentHeat is NaN.
01-Adir-01:003.829(Severe) Heating : [Heritage] H19 - activeVentHeat is NaN.
01-Adir-01:003.829(Severe) Heating : [Heritage] H16 - activeVentHeat is NaN.
01-Adir-01:003.829(Severe) Heating : [Heritage] H15 - activeVentHeat is NaN.
```
The heat calculation is bad because the area of a hallway has variable length. 

Therefore, the area of a hallway cannot be calculated in BuildingConfig when reading buildings.xml.

But the length of a hallway is actually defined in a settlement template. 

For instance, in settlement_phase_3-sx.xml, it will show each hallway's length.

I'll fix this.

Issue #1729, opened 2025-09-06
Unit classes have too many interfaces 


bevans2000:
## Problem

There are a number of inventory management interfaces, e.g. ```ResourceHolder``` that define the responsibility of an object to manage inventory. Logic elsewhere to manage inventory always uses the interfaces, e.g. ```LoadingController```.

The original decision was to implement the various interfaces with concrete classes, e.g. ```EquipmentInventory```  and then use a proxy pattern to expose these at the Unit level. This results in a lot of proxy boilerplate methods in the Unit sub-classes.

This has proved to be problematic when making changes as it increases the impact of any change to many classes. Plus gives the Unit classes a bloated feeling.

## Solution

The proposed solution is to drop the inventory interfaces from the Units which will significantly reduce the coupling and number of boilerplate methods. The Units will instead provider getters to the underlying concrete classes returning the appropriate interface, emg ```getBinHolder```. The logic manipulating the inventory will have to call the appropriate getters before executing the existing logic.

This change can be done using a gradual approach by converting the inventory using logic gradually. Then removing the interfaces at the Unit level at the end.

There is a risk that a valid use of instanceof may hide a conversion point that could be missed but these can be caught via testing.

## Supporting Changes

EquipmentInventory should extend MicroInventory instead of delegation
EquipmentOwner interface should implement ItemHolder

bevans2000:
Yes I see no reason why we can not do this as one interface at a time.

bevans2000:
This is just a big refactor exercise that we can use GitHub Copilot for.

mokun:
> a big refactor exercise that we can use GitHub Copilot for.

btw, I heard are they charging tokens on those [GitHub Actions](https://github.com/mars-sim/mars-sim/actions).

The old ones are heaping up.

Can we turn them off in yml files since they probably are not needed ?

bevans2000:
I can use CoPilot as I have a subscription. If I ask CoPilot to resolve an issue, it will be under my subscription.

mokun:
So will this be done soon ?

bevans2000:
Maybe, I wanted us to get the Visit Landmark Mission done first

bevans2000:
Plus I want to complete the migration to a single table model design.

bevans2000:
Need to remove getSuitSet, should use the single getEquipmentSet and an overloaded method that takes EquipmentType argumetn to filter.

bevans2000:
[...]

PR #2051, opened 2026-08-15
Task: Use Settlement Equipment Inventory instead of direct access

bevans2000:
This is the next stage in the #1729 changes. It focuses on the use of Settlement in resource management.
There are 2 high level changes in this PR:
- UnitTests change to use the `Settlement.getEquipmentInventory` method instead of direct access
- In the core logic use the getEquipmentInventory


mokun (approved):

mokun:
Can you remerge with the latest commits ? 

The unit tests are fixed now.

PR #2072, opened 2026-08-31
feat: Convert Exploration mission to new MissionProject

bevans2000:
Key changes:
- Created a new class EVAMissionStep to represent a step in an EVA mission.
- EVAMissionStep will wait for all EVAs to be completed before moving to the next step in the mission.
- Modified VisitLandmark to use common EVA logic
- Create ExploreSiteStep to bind an ExploreSite task to a Mission
- ExplorationMeta builds a MissionProject with steps
- Delete old Exploreatino mission class

Also fixes the bugs:
- Updated Mind to fix problem with missions being skipped

close #975

PR #2091, opened 2026-09-12
fix(core): Evicted Worker leave mission

bevans2000:
Changes:
- MissionBoardVehicleStep releases the evicted member from the Mission
- Add UnitTest for board vehicle step
- MissionBoardVehicleStepTest provides a reusable method createRoster
- ExplorationMetaTest uses new createRoster method

close #2072

### DIFF
--- a/mars-sim-core/src/main/java/com/mars_sim/core/mission/steps/MissionBoardVehicleStep.java
+++ b/mars-sim-core/src/main/java/com/mars_sim/core/mission/steps/MissionBoardVehicleStep.java
@@ -39,7 +39,7 @@ public class MissionBoardVehicleStep extends MissionStep {
 	private static final SimLogger logger = SimLogger.getLogger(MissionBoardVehicleStep.class.getName());
 
 	/* How long do Worker have to complete departure */
-	private static final int DEPARTURE_DURATION = 150;
+	protected static final int DEPARTURE_DURATION = 150;
 	private static final int DEPARTURE_PREPARATION = 15;
 
     private boolean notifiedMembers = false;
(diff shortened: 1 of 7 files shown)


mars-sim-core/src/main/java/com/mars_sim/core/mission/steps/MissionBoardVehicleStep.java
```diff
@@ -139,7 +139,7 @@ private boolean evictStragglers(MissionProject m, Vehicle v) {
 				&& ((members.size() - ejectedMembers.size()) >= 2)) {
 			for(Person ej : ejectedMembers) {
 				logger.info(ej, "Ejected from mission " + m.getName() + " missed Departure");
-				m.removeMember(ej);
+				ej.setMission(null);
```
mokun, line 142: In AbstractMission,
```java
/**
	 * A Member leaves the Mission and adjust his work shift.
	 * Note: should use removeMember() to call memberLeave().
	 */
	private final void memberLeave(Worker worker) {
		// Added codes in reassigning a work shift
		if (worker instanceof Person person) {
			
			logger.info(person, "Removed from " + worker.getMission() + ".");
			
			String buildingName = worker.getBuildingLocation() != null ? worker.getBuildingLocation().getName() : null;
			worker.getTaskManager().recordActivity(getName(), getPhaseDescription(), "Leaving mission", getName(), buildingName);
			
			worker.setMission(null);
			
	      	if (RoleType.GUEST != person.getRole().getType() && person.getShiftSlot() != null) {      
	      		person.getShiftSlot().setOnCall(false);
	      	}	

			registerHistoricalEvent(person, HistoricalEventType.MISSION_FINISH, "Removing a member");
			fireMissionUpdate(REMOVE_MEMBER_EVENT, worker);
		}
	}
```

Shall we by any chance also cover these basis when calling setMission(null) in the new mission project paradigm ?

bevans2000: The event is handles in the new mission code. Changing the shift is the responsibility of the Person and it should go there not on the Mission

PR #2092, opened 2026-09-12
task(tools): Create Process Check tool

### LINKED ISSUE #2086: Create unit tests to check if the mass of input resources equals that of output resources
- Currently, we have to go back to each process in `manufacturing.xml` and `resource_process.xml` to double check if the mass of input resources equals that of output resources.
- In other words, it makes sense to follow through on the conservation of mass in each process to ensure the equation is balanced.
- An additional advantage is that this will help correcting the predefined mass of each part in parts.xml.

- Write a method as an unit test that prints out the results 

**Describe alternatives you've considered**
- Write a main method that can be run independently to show any processes that need to be fixed, without having to bother with unit tests.

**Version**
- pre-4.1

**Notes**
- It will need to read in all the manufacturing process before it can compute the total input mass and the total output mass because the assembly of some derived amount resource, part, equipment, robot, vehicle, etc are found in manufacturing.xml.
- It will need to recursively compute each part or each equipment that may comprise other parts and resources.
- By looking up the input resources that create a part, it will check if the stated mass of each part is indeed correct.

bevans2000:
This adds a new tool, ProcessCheck, for the Mars simulation environment. It check that the total output mass is not larger than teh total input mass of process specs. It is applicable to Salvage, Food & Manufacturing.

This is a help file `process-check.md` to explain the operation.

close #2086

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,6 @@
 .DS_Store
 lib/
 **/*.db
+
+# ProcessCheck output
+process-check-details.txt
--- a/mars-sim-tools/process-check.md
+++ b/mars-sim-tools/process-check.md
@@ -1,0 +1,50 @@
+# ProcessCheck
+
+ProcessCheck validates process definitions.
+It scans the configured manufacture, food production, and salvage processes and flags any process where total output mass is greater than total input mass.
+The report is grouped by process type.
+
+## Output
+
+- Standard output: grouped summary of failed processes by process type.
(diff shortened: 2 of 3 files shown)


mokun (approved):

mokun:
Awesome !

bevans2000:
> Awesome !

Yes, it's good. Liked what CoPilot did. The instruction file explain how to run it from Maven but of course you can run it direct from the IDE as well. Give it a try, just merging

PR #2095, opened 2026-09-12
fix: Change the InviteStudyCollaboratorTest to reduce failures

bevans2000:
Changes:
- InviteStudyCollaborator exposes invitee list
- Test checks invitee list size
- Test modifies relationship opinion the primary researcher has of the collaborators
- Reduce excessive warning output when placing initial people


mars-sim-core/src/test/java/com/mars_sim/core/science/task/InviteStudyCollaboratorTest.java
```diff
@@ -85,6 +86,11 @@ public void testCreateTask() {
         BuildingManager.addToActivitySpot(c1, l, FunctionType.RESEARCH);
         BuildingManager.addToActivitySpot(c2, l, FunctionType.RESEARCH);
 
+        // Make primary like others
+        RelationshipUtil.changeOpinion(p, c1, 100);
```
mokun, line 90: There you go !

mars-sim-core/src/main/java/com/mars_sim/core/building/BuildingManager.java
```diff
@@ -1100,10 +1100,6 @@ public static boolean addPersonToBuildingSpotByJobType(Person person, Settlement
 			}
 		}
 
-		if (!found) {
```
mokun, line 1103: Feel like we could dedicate some spots to be a "collaborative" workspace. That means they are for multi-purposes. 

Some places in the U.S. (before Covid) champion this idea. It can be a building where many small companies rent dozens of workspace out. 

https://www.coworkingcafe.com/blog/15-unique-us-coworking-spaces/

mokun (approved):

mars-sim-core/src/test/java/com/mars_sim/core/science/task/InviteStudyCollaboratorTest.java
bevans2000: Not 100% sure but it should help

Issue #2077, opened 2026-09-05
Settlement Map improvements


bevans2000:
**Describe the improvement**
The recent addition of the data collection entity has shown the Settlement map panel is not easily expandable as other layers are added. Adding a new Entity to display involves a lot of code changes in numerous classes.

More importantly, the current logic is not efficient when searching for mouse actions on the rendered items as it scans and draws all potential entities independently of the actual rendered image. So for a large Settlemetn everything is drawn but less than 10% is visible to the user.

** Solution **
The solution is to follow the hotspot pattern of the Navigator panel where the layers return hotspots that represent rendered "things".  The hotspot would identify a region and what actions/details are available. It removes the need for specific coding within the central Settlement Map Panel. Essentially implementing a Command pattern.

Secondly, the hotspots returned would only contain those that have been rendered, not the whole potential scope. The layer must also be able to identify Entities not visible and excluded from rendering. Could this be done as a rough distance from the center of the display with a tolerance.

Furthermore, the means to externally select Entities can be centralized around a single Entity instead of methods for each type. Only one Entity is highlighted in the map currently but it uses complex logic to achieve this.

Lastly, filter options should be defined by the specific layer using that filter instead of central.
 .

bevans2000:
@mokun What do you think about removing the weather overlay from the Settlement Map? The information is available elsewhere and does the user want that information when looking at the map?
Same for the ticker bar message. I don't use it. Both just take up space

PR #2107, opened 2026-09-19
task(ui): Highlight selected Entity when popup menu shown

bevans2000:
Changes:
- SettlementMapPanel updates the selected Entity when target of popup
- Settlement manages parked vehicles & indoor citizens as concurrent sets as large read with background updates

Ref #2077


bevans2000:
@mokun I have had to change the DataCollectionSite class in this PR. It uses a LocalPosition but never records within which Settlement the position is within. Anything that is positioned in a Settlement must now implement a SettlementPOI

mars-sim-core/src/main/java/com/mars_sim/core/structure/Settlement.java
```diff
@@ -432,7 +432,7 @@ protected Settlement(String name, Coordinates location) {
 		robotsWithin = new UnitSet<>();
 		deathRegistry = new UnitSet<>();
 
-		// This are read many and write infrequesnt so suitable for concurrent collections
+		// These collections are read many and write infrequent so suitable for concurrent collections
```
mokun, line 435: Aren't they are used relatively frequently ? There are people and vehicle going in and out of a settlement all the time.

The best way to avoid CME in VehicleMapLayer is to use one single flag  isReady in a settlement to indicate that a vehicle has just been removed from parkedNGaragedVehicles Set when it's about to depart a settlement.

This effectively eliminates CME without having to rely on CopyOnWriteArraySet that uses more CPU for the default scenario of 19 settlements.

After this PR is done, is it okay that if I demonstrate isReady flag is good, we can switch back to using it ?

mokun (approved):

mars-sim-core/src/main/java/com/mars_sim/core/structure/Settlement.java
bevans2000: The ready flag doesn't work. It does not prevent a list being scanned when there is a parallel change occurring. We scan the Person's indoors and parked vehicle list numerous places, these are all cases of a potential CME.
The drawback is not CPU, there is a small extra memory used during the update operation. If you read the blogs this is a very efficient copy of the underlying byte array.
Let's not add it back in, this is cleaner and easy to understand.

Issue #2027, opened 2026-07-25
Create dynamic inventory that optimises EntityEvents & Metrics


bevans2000:
## Problem

The current Resource inventory classes generate a lot of EntityEvents when a resource or part changes. This can results in millions of events per minute on the default scenario. In particular, if a resource changes multiple times in the same pulse, multiple events will be created.
This has a knock on impact for the Metric manager since if each change resulted in a new data point it would create millions.
The number of "recorded" changes need to be reduced to just one per Pulse per Resource per Settlement.
    
## Solution

The solution proposes to react to changes only once per Pulse instead of immediately. This means "remembering" the changes each pulse. This can eb done as part of the MicroInventory class.

Create a sub class of the MicroInventory class (or change that class) such that EntityEvent & Metrics creations are collapsed. The resources decrease/increase multiple time in a single pulse but only want a single record per pulse.

This idea is to only create EntityEvent & Metrics once per Pulse by caching the changes for the duration of a single pulse. The idea is to hold a transient cache of changed resources and then flush the cache at the end of the end of the Pulse via a new method, i.e. MicroInventory would implement the Temporal interface.

Also combine Items and Amount resources into one metric/event.

In addition need to have a global concept of trackable resources so not to track everything in the Metric Manager everything. Should be configurable.

## Code
```
// Record a change in a single pulse
private class ResourceChange {
    double consumed;
    double stored;
    void addStored(value) {
         stored += value;
   }
}

private transient Map<Integer,ResoruceChange> changes; 

// Modified store method pattern
public storeXXXX(resourceId, value)
  changes.computeIfAbsent(resourceId, new ResourceChange()).addStored(value);
  .....
}

// Fires once per pulse and flushes the cahce
public void clockPulse(pulse) {
     for(var entry : changes) {
         var resId = entry.getKey();
         if (trackedResources.contained(resId)) {
             if (consumed > 0)
                resourceMetrics.add("Consumed", entry.getKey, entry.getValue().consumed);
             if (stored > 0) 
              resource....
         }
         fireEntityEvent(INVENTORY_CHAGE, entry.getKey);     
   }

   // Reset cache
   changed = new HashMap();
}
```

mokun:
> The number of "recorded" changes need to be reduced to just one per Pulse per Resource per Settlement.

Yes this micro-cache is a long awaited feature that we could have it since the beginning.

[...]

PR #2109, opened 2026-09-19
task(core): Rework the MicroInventory class

bevans2000:
This PR does the prep and clean up work to enable the management of events and metrics as part of #2027. It completes:

- EquipmentInventory extends MicroInventory to reduce excessive delegation code
- Building no longer implements Resource and Item holder but use parent inventory
- Redesign of stock storage in MicroInventory so only a single list is needed


mokun:
Very nice for cleaning up the micro inventory class.

mokun:
Will this PR or the next PR take care of consolidating all the small little store and retrieve actions from all directions into one big store and retrieve action in one frame ?

bevans2000:
Yes this will generate one event per frame for each changes resource. All those little changes generate a lot of events.
This pattern will allow it to add metrics to the metrics manager as well

bevans2000:
@mokun I want to create a new Tool Window called Performance. It will show the output of the EventManager stats so we can observe the load Events are creating when running a simulation. Also I will move the Pulse metrics from the Time Tool into this new tool. This pulse metrics are all about performance of the simulation, not the time.

mokun:
> This pulse metrics are all about performance of the simulation, not the time

Yes that statement makes sense !

But I would love to have the MasterClock send out pulses in a regulated fashion so that the simulation remains stable, especially when we increase the speed of the simulation.

It means that even if we increase it to a higher time ratio than before, the pulse width of each pulse become larger. It would not impact the bulk of the simulation.

bevans2000:
I'm just moving where they are displayed, not changing them

mokun:
> Yes this will generate one event per frame for each changes resource. 

So, is it true you're essentially create a cache in each EquipmentInventory's MicroInventory to let all the incoming store actions and all the outgoing retrieval actions cancel out each other and calculate the balance for one action.

What if this cache comes out negative at the end, say, it's -24 kg of oxygen. And say the MicroInventory own have 23 kg of oxygen ? Then you have to look up the record and decide which one retrieval and subtract that ?

>  create a new Tool Window called Performance. 

So, will this tool also capture this cache hit rate of the use of MicroInventory cache ?

Maybe we could also track the cache hit of the Mars surface coordinates ?!

bevans2000:
I haven't actually done it yet ,🤣🤣
But it's not a cache approach, the same logic is used tobtrabk ins and outs. But they is a flag set that record any resources changed during a frame, then at the end of the frame those resources generate a single event each. Quite simple really. The flag set is reset each frame
Then further those resources changed can be added to the Metric Manager

